PR #4, opened 2019-05-03
update readme to be more userfacing

slopp:
We are working on making these two repos public:

https://github.com/rstudio/r-docker
https://github.com/rstudio/r-system-requirements

I'd like to document the whole hirearchy:
r-builds > r-docker > r-system-requirements

This PR updates the README for this repo to add some more context for users. Please review when possible, and then we can merge when those other two repos (now referenced here) go public.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,69 @@
 # r-builds
 
-The environment which produces R builds obtainable from: https://cdn.rstudio.com/r/versions.json
+This repository orchestrates tools to produce R binaries obtainable from:
+https://cdn.rstudio.com/r/versions.json. The binaries are available as a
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,69 @@
...
+
+## Example Usage
+
+These binaries are designed to be copied onto a server, as opposed to installed
+using a system package manager like `apt` or `yum`. This approach allows administrators
+to offer multiple versions of R side-by-side. 
+
+```
+# Pre-req install runtime pre-reqs
+# Then copy the desired R version from the CD
```
glin: I would just say "Install runtime pre-reqs"

CD -> CDN

PR #5, opened 2019-05-03
Fix default pager for CentOS and openSUSE builds

### LINKED ISSUE #3: Default pager is specific to Ubuntu/Debian
On CentOS and openSUSE, help pages and `file.show()` fail because `/usr/bin/pager` doesn't exist:
```sh
# R 3.5.3 on CentOS 7

> ?print
/opt/R/3.5.3/lib/R/bin/pager: line 11: /usr/bin/pager: No such file or directory
Warning message:
error in running command 

> file.show("test.txt")
/opt/R/3.5.3/lib/R/bin/pager: line 11: /usr/bin/pager: No such file or directory
Warning message:
error in running command
```

I think it's because the default pager is configured as `/usr/bin/pager`, which seems to be an Ubuntu/Debian specific thing. https://unix.stackexchange.com/questions/5383/which-systems-have-pager-shortcut-alias/5404#5404
https://github.com/rstudio/r-builds/blob/b822b418a0b9d7db9b2ba418d3bc3bf35107463f/builder/build.sh#L76

---

From the R docs, if `PAGER` isn't set, R defaults to `less` if available and `more` as a fallback:
> The ‘factory-fresh’ default is to use ‘R_HOME/bin/pager’, which is a shell script running the command-line specified by the environment variable PAGER whose default is set at configuration, usually to less. On a Unix-alike more is used if pager is empty.
> https://stat.ethz.ch/R-manual/R-devel/library/base/html/options.html

So I think we could fix this by setting `PAGER=/usr/bin/pager` for Ubuntu and Debian builds only, and letting CentOS/openSUSE builds use the default. It should be `more` on the openSUSE images and `less` on CentOS images (but we may want to remove `less` as a dependency).

glin:
Fixes #3

This makes the `PAGER` override apply to Ubuntu/Debian builds only. The configure script should now pick up valid pagers for the other OSs, which will either be `less` or `more`.

I also removed the `less` dependency for CentOS so it doesn't become a requirement to use R.

### DIFF
--- a/builder/Dockerfile.centos-6
+++ b/builder/Dockerfile.centos-6
@@ -17,7 +17,6 @@ RUN yum -y update \
     krb5-devel \
     krb5-libs \
     lapack-devel \
-    less \
     libICE-devel \
     libSM-devel \
     libX11-devel \
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -16,7 +16,6 @@ RUN yum -y update \
     java-1.8.0-openjdk-devel \
     java-1.8.0-openjdk-headless \
     lapack-devel \
-    less \
     libICE-devel \
     libSM-devel \
     libX11-devel \
--- a/builder/Dockerfile.debian-9
+++ b/builder/Dockerfile.debian-9
@@ -13,5 +13,8 @@ RUN pip install awscli
 
 RUN chmod 0777 /opt
 
+# Override the default pager used by R
+ENV PAGER /usr/bin/pager
+
 COPY build.sh .
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.ubuntu-1604
+++ b/builder/Dockerfile.ubuntu-1604
@@ -13,5 +13,8 @@ RUN pip install awscli
 
 RUN chmod 0777 /opt
 
+# Override the default pager used by R
+ENV PAGER /usr/bin/pager
+
 COPY build.sh .
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.ubuntu-1804
+++ b/builder/Dockerfile.ubuntu-1804
@@ -13,5 +13,8 @@ RUN pip install awscli
 
 RUN chmod 0777 /opt
 
+# Override the default pager used by R
+ENV PAGER /usr/bin/pager
+
 COPY build.sh .
 ENTRYPOINT ./build.sh
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -73,7 +73,6 @@ compile_r() {
   # set some common environment variables for the configure step
   AWK=/usr/bin/awk \
   LIBnn=lib \
-  PAGER=/usr/bin/pager \
   PERL=/usr/bin/perl \
   R_PDFVIEWER=xdg-open \
   R_BROWSER=xdg-open \


glin:
@stevenolen yeah sorry, I should've included some testing notes. I built R 3.1.3 and 3.6.0 on each OS ( CentOS 6/7, Ubuntu 16/18, Debian 9, openSUSE 42/15), checked the default `PAGER`, and tried to open a help page.

```sh
docker build -t rstudio/r-builds:centos-6 --file builder/Dockerfile.centos-6 builder
docker run -it --rm --entrypoint "/bin/bash" rstudio/r-builds:centos-6
###
# build R
R_VERSION=3.6.0 ./build.sh
# open help page
/opt/R/*/bin/R -e '?print'
# get default pager
/opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
```

Here are the default pagers for each OS. They're all `more` on CentOS/openSUSE except CentOS 6, which uses `less`. But I checked the base `centos:6` image and found that it does indeed ship with `less`.
```sh
# CentOS 6
[root@bdd64ec42399 /]# /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/less"
[root@bdd64ec42399 /]# cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/less'}

# CentOS 7
[root@c00dbd27cfbb /]# /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/more"
[root@c00dbd27cfbb /]# cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/more'}

# openSUSE 42
2399dc60821b:/ # /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/more"
2399dc60821b:/ # cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/more'}

# openSUSE 15
77daeadf9603:/ # /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/more"
77daeadf9603:/ # cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/more'}

# Ubuntu 16
root@d93eb876b77e:/# /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/pager"
root@d93eb876b77e:/# cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/pager'}

# Ubuntu 18
root@90a30498229a:/# /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/pager"
root@90a30498229a:/# cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/pager'}

# Debian 9
root@6d4f2775b29d:/# /opt/R/*/bin/Rscript -e 'Sys.getenv("PAGER")'
[1] "/usr/bin/pager"
root@6d4f2775b29d:/# cat /opt/R/*/lib*/R/etc/Renviron | grep PAGER
PAGER=${PAGER-'/usr/bin/pager'}
```

stevenolen:
builds appear to have finished, after incorporating this PR!

PR #12, opened 2019-07-02
Allows building deb packages for R on ubuntu-1804

sellorm:
This update provides a mechanism for building deb packages on Ubuntu
1804. Several changes have been necessary to facilitate this:

* Dockerfile has been updated. It now includes:
    - fpm - a tool tool to create OS specific packages
    - copying the package script into the image
* the OS specific package.ubuntu-1804 packaging script
    - this is the script that uses fpm to build the package
* Updates to build.sh. It now includes:
    - runs the package script if it exists
    - stores the resultant package to s3

*Note:* upload to s3 has not been tested.

### DIFF
--- a/builder/Dockerfile.ubuntu-1804
+++ b/builder/Dockerfile.ubuntu-1804
@@ -11,10 +11,15 @@ RUN set -x \
 
 RUN pip install awscli
 
+RUN apt-get install -y gcc make ruby ruby-dev wget
+
+RUN gem install fpm
+
 RUN chmod 0777 /opt
 
 # Override the default pager used by R
 ENV PAGER /usr/bin/pager
 
+COPY package.ubuntu-1804 /package.sh
 COPY build.sh .
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.ubuntu-1804
+++ b/builder/Dockerfile.ubuntu-1804
@@ -6,12 +6,10 @@ RUN set -x \
   && sed -i "s|# deb-src|deb-src|g" /etc/apt/sources.list \
   && export DEBIAN_FRONTEND=noninteractive \
   && apt-get update \
-  && apt-get install -y libcurl4-openssl-dev libicu-dev libopenblas-base wget python-pip \
+  && apt-get install -y libcurl4-openssl-dev libicu-dev libopenblas-base wget python-pip ruby ruby-dev \
   && apt-get build-dep -y r-base
 
 RUN pip install awscli
-
-RUN apt-get install -y gcc make ruby ruby-dev wget
 
 RUN gem install fpm
 
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -19,6 +19,12 @@ upload_r() {
   if [ -n "$S3_BUCKET" ] && [ "$S3_BUCKET" != "" ]; then
     echo "Storing artifact on s3: ${S3_BUCKET}, tarball: ${TARBALL_NAME}"
     aws s3 cp /tmp/${TARBALL_NAME} s3://${S3_BUCKET}/${S3_BUCKET_PREFIX}${baseName}/${TARBALL_NAME}
+    # check if PKG_FILE has been set by a packager script and act accordingly
+    if [ -n "$PKG_FILE" ] && [ "$PKG_FILE" != "" ]; then
+      if [ -f "$PKG_FILE" ]; then
+	aws s3 cp ${PKG_FILE} s3://${S3_BUCKET}/${S3_BUCKET_PREFIX}${baseName}/pkgs/$(basename ${PKG_FILE})
+      fi
+    fi
   fi
   if [ -n "$LOCAL_STORE" ] && [ "$LOCAL_STORE" != '' ]; then
     echo "Storing artifact locally: ${LOCAL_STORE}, tarball: ${TARBALL_NAME}"
@@ -89,6 +95,16 @@ compile_r() {
   make install
 }
 
+# check for packager script
+## If it exists this build is ready for packaging with fpm, so run the script
+## else do nothing
+package_r() {
+  if [[ -f /package.sh ]]; then
+    export R_VERSION=${1}
+    source /package.sh
+  fi
+}
+
 set_up_environment() {
   mkdir -p /opt/R
 }
@@ -101,6 +117,7 @@ _version_is_greater_than() {
 set_up_environment
 fetch_r_source $R_VERSION
 compile_r $R_VERSION
+package_r $R_VERSION
 clean_r
 archive_r $R_VERSION
 upload_r $R_VERSION
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -117,7 +117,7 @@ _version_is_greater_than() {
 set_up_environment
 fetch_r_source $R_VERSION
 compile_r $R_VERSION
-package_r $R_VERSION
 clean_r
+package_r $R_VERSION
 archive_r $R_VERSION
 upload_r $R_VERSION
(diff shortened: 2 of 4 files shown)


builder/build.sh
```diff
@@ -101,6 +117,7 @@ _version_is_greater_than() {
 set_up_environment
 fetch_r_source $R_VERSION
 compile_r $R_VERSION
+package_r $R_VERSION
```
glin: I think `clean_r` runs some special install steps for CentOS 6 specifically, so it should come before the packaging step (@jonyoder can confirm).

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,50 @@
+if [[ ! -d /tmp/output/ubuntu-1804 ]]; then
+  mkdir /tmp/output/ubuntu-1804
```
glin: I got a "mkdir: cannot create directory" error when I was testing this out at first, and had to add a `-p` flag here

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,50 @@
...
+  -n R-${R_VERSION} \
+  -p /tmp/output/ubuntu-1804/ \
+  --prefix=/opt/R \
+  -d zip \
+  -d unzip \
+  -d libpaper-utils \
+  -d xdg-utils \
+  -d libblas3 \
+  -d libatlas3-base \
+  -d libblas3 \
```
glin: Some of these packages seem to be duplicated? But now that we have OpenBLAS, we could remove these other BLAS/LAPACK libraries - libatlas, libblas, and liblapack.

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,50 @@
+if [[ ! -d /tmp/output/ubuntu-1804 ]]; then
+  mkdir /tmp/output/ubuntu-1804
+fi
+
+fpm \
```
glin: I noticed that the deb package had a priority of extra, which is the fpm default, but also deprecated apparently. Should we set it to "optional"? https://github.com/jordansissel/fpm/issues/1398

glin: Do we need to add package metadata like license, description, maintainer? If it's coming later, feel free to ignore

builder/build.sh
jonyoder: Yes, I would agree. The `clean_r` step is really important for CentOS.

builder/package.ubuntu-1804
sellorm: ✅ Done - I've used the repo url in place of the "Maintainer"

sellorm: ✅ Done - good catch

builder/package.ubuntu-1804
sellorm: ✅ Done

builder/package.ubuntu-1804
sellorm: ✅ Done

builder/build.sh
sellorm: ✅ Done - changed the order of the steps

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,49 @@
...
+  -d libopenblas-base \
+  -d libbz2-1.0 \
+  -d libc6 \
+  -d libcairo2 \
+  -d libcurl4 \
+  -d libglib2.0-0 \
+  -d libgomp1 \
+  -d libicu60 \
+  -d libjpeg8 \
+  -d liblapack3 \
```
glin: This liblapack3 can be removed as well

sellorm: Well spotted - can't believe I missed that from the original commit :man_facepalming: 

The idea - at least as far as it's implemented here is that docker-compose picks up the BUILD_NUMBER from Jenkins and passes it through to the container as BUILD_NO. The packaging script picks it up from there.

I've just pushed a new commit for the docker-compose.yml that has the BUILD_NO/BUILD_NUMBER bits in it for all OSes.

sellorm: I've also pushed another commit which removes liblapack3 per your request, @glin.

builder/Dockerfile.ubuntu-1804
```diff
@@ -11,10 +11,15 @@ RUN set -x \
 
 RUN pip install awscli
 
+RUN apt-get install -y gcc make ruby ruby-dev wget
```
stevenolen: include these in the `apt-get install` line above so that they are in the same layer

builder/docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
   ubuntu-1604:
     command: ./build.sh
     environment:
+      - BUILD_NO=${BUILD_NUMBER}
```
stevenolen: The docker-compose environment is only used for local dev assistance, and the actual build aren't related to a jenkins build in any way (and so don't have a `BUILD_NUMBER` to inherit). Also, it would be best to keep the variable name the same, rather than using `BUILD_NO` since "no" can have a different meaning than "number" and we aren't really trying to save bytes 😉

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,48 @@
...
+  -v ${BUILD_NO} \
+  -n R-${R_VERSION} \
+  --deb-priority "optional" \
+  --deb-field 'Bugs: https://github.com/rstudio/r-builds/issues' \
+  --url "http://www.r-project.org/" \
+  --description "GNU R statistical computation and graphics system" \
+  --maintainer "RStudio Inc https://github.com/rstudio/r-builds" \
+  --license "GPL-2" \
+  -p /tmp/output/ubuntu-1804/ \
+  -d zip \
```
jspiewak: Sort alphabetically so it is easy to find things.

builder/package.ubuntu-1804
```diff
@@ -0,0 +1,48 @@
...
+  -d unzip \
+  -d libpaper-utils \
+  -d xdg-utils \
+  -d libopenblas-base \
```
jspiewak: I thought we weren't going to build against a particular BLAS implementation?

builder/docker-compose.yml
sellorm: I’ll change the BUILD_NO thing, that’s easy enough to fix.

The packaging script currently uses the build number in place of a software version.

In order to ensure we can still install these packages side-by-side, I’ve moved the official R version number into the packages name and am then using the BUILD_NUMBER as the version number of the package.

So packages comes out with names like:

R-3.6.0-123-amd64.deb

Where ‘r-3.6.0’ is the package name and ‘123’ is the package version. The idea was to allow for updates where the R version hasn’t changed, but maybe we’ve improved the build or packaging processes and re-created the deb/rpm.

If we can’t use the Jenkins BUILD_NUMBER, is there another number we could use in it’s place??

builder/package.ubuntu-1804
sellorm: I'm not 100% certain what the situation is around that - I'll try to clarify.

builder/Dockerfile.ubuntu-1804
sellorm: Done

builder/docker-compose.yml
sellorm: Removed the reliance on Jenkins BUILD_NUMBER completely - all built packages are now created with version 1

builder/package.ubuntu-1804
sellorm: Done

builder/package.ubuntu-1804
sellorm: Confirmed libopenblas-base not required by current builds and removed

PR #14, opened 2019-07-05
Update openSUSE base image

glin:
Switch from the deprecated `opensuse:42.3` to `opensuse/leap:42.3`

### DIFF
--- a/builder/Dockerfile.opensuse-42
+++ b/builder/Dockerfile.opensuse-42
@@ -1,4 +1,4 @@
-FROM opensuse:42.3
+FROM opensuse/leap:42.3
 
 ENV OS_IDENTIFIER opensuse-42
 


tylfin:
@glin Probably okay here, but have we ever considered the security ramifications of pulling public docker images? We could use a security scanner like https://coreos.com/clair/docs/latest/ to ensure secure images and push them to a private repository

trestletech:
That's a great idea. I'm unaware of any prior conversations but I'd love for that to be a part of our "promoted R docker images."

My guess is that @tylfin or @kevinhankens could coordinate with the Hosted team to settle on a strategy that we might be able to recycle across the company. I'd defer to @slopp in terms of prioritization, though.

slopp:
Hmmm this seems like a much broader topic, starting from an OS level image is ubiquitous from what I've seen. 

I think it'd be an interesting question to run by our security team / Joe and others, but I wouldn't pull the fire alarm.

PR #20, opened 2019-07-28
Enables output of Linux OS packages for CentOS 7

sellorm:
Builds on earlier work and provides the tooling required to create Linux OS packages (deb in this case).

**Modifies**: Dockerfile.centos-7

So that it includes the fpm package build tool and also copies in the packaging script

**Adds**: package.centos-7

which is the packaging script which runs fpm with the required parameters and dependencies

### DIFF
--- a/builder/package.centos-7
+++ b/builder/package.centos-7
@@ -38,12 +38,10 @@ fpm \
   -d pcre-devel \
   -d pcre2-devel \
   -d perl-Digest-MD5 \
-  -d sudo \
   -d tcl \
   -d tk \
   -d tre \
-  -d vim \
-  -d wget \
+  -d which \
   -d xz-devel \
   -d zlib-devel \
   /opt/R/${R_VERSION}
--- a/builder/package.centos-7
+++ b/builder/package.centos-7
@@ -2,6 +2,12 @@ if [[ ! -d /tmp/output/centos-7 ]]; then
   mkdir -p /tmp/output/centos-7
 fi
 
+# create post-install script required for openblas
+cat <<EOF >> /post-install.sh
+mv /opt/R/${R_VERSION}/lib/R/lib/libRblas.so /opt/R/${R_VERSION}/lib/R/lib/libRblas.so.keep 
+ln -s /usr/lib64/libopenblasp.so /opt/R/${R_VERSION}/lib/R/lib/libRblas.so
+EOF
+
 fpm \
   -s dir \
   -t rpm \
@@ -14,6 +20,7 @@ fpm \
   --description "GNU R statistical computation and graphics system" \
   --maintainer "RStudio Inc https://github.com/rstudio/r-builds" \
   --license "GPL-2" \
+  --after-install /post-install.sh \
   -p /tmp/output/centos-7/ \
   -d bzip2-devel \
   -d gcc \
--- a/builder/package.centos-7
+++ b/builder/package.centos-7
@@ -1,0 +1,46 @@
+if [[ ! -d /tmp/output/centos-7 ]]; then
+  mkdir -p /tmp/output/centos-7
+fi
+
+fpm \
+  -s dir \
+  -t rpm \
+  -v 1 \
+  -n R-${R_VERSION} \
+  --vendor "R Project" \
+  --deb-priority "optional" \
+  --deb-field 'Bugs: https://github.com/rstudio/r-builds/issues' \
+  --url "http://www.r-project.org/" \
+  --description "GNU R statistical computation and graphics system" \
+  --maintainer "RStudio Inc https://github.com/rstudio/r-builds" \
+  --license "GPL-2" \
+  -p /tmp/output/centos-7/ \
+  -d bzip2-devel \
+  -d gcc \
+  -d gcc-c++ \
+  -d gcc-gfortran \
+  -d libcurl-devel \
+  -d libicu-devel \
+  -d libSM \
+  -d libtiff \
+  -d libXmu \
+  -d libXt \
+  -d make \
+  -d openblas-devel \
+  -d pango \
+  -d pcre-devel \
+  -d pcre2-devel \
+  -d perl-Digest-MD5 \
+  -d sudo \
+  -d tcl \
+  -d tk \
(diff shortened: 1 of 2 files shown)


builder/package.centos-7
```diff
@@ -0,0 +1,46 @@
...
+  -d libtiff \
+  -d libXmu \
+  -d libXt \
+  -d make \
+  -d openblas-devel \
+  -d pango \
+  -d pcre-devel \
+  -d pcre2-devel \
+  -d perl-Digest-MD5 \
+  -d sudo \
```
glin: I think we should remove sudo, and maybe vim/wget. The latter two could be useful, but we didn't add them for the Ubuntu package.

builder/package.centos-7
```diff
@@ -0,0 +1,46 @@
...
+  -d gcc-c++ \
+  -d gcc-gfortran \
+  -d libcurl-devel \
+  -d libicu-devel \
+  -d libSM \
+  -d libtiff \
+  -d libXmu \
+  -d libXt \
+  -d make \
+  -d openblas-devel \
```
glin: To use OpenBLAS with R on CentOS and openSUSE, there's an additional step of replacing the default BLAS with OpenBLAS. This is what we do in the CentOS 6/7 images: https://github.com/rstudio/r-docker/blob/c7e58d0be50259c7047a9f4eeb6da562ae3f7175/Dockerfile-centos.template#L16-L18

Can we do this automatically when users install the RPM, like as a post-install step?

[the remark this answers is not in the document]
sellorm: There isn't a sensible way to work around this unfortunately. Ultimately the dependency is on those packages, rather than EPEL, so in situations where EPEL is not permitted users can still manually download those packages and install them first. I think this is as good as it's going to get, sadly.

sellorm: Ah, I didn't realise. We can add a post install step though, so I'll take a look at doing that.

builder/package.centos-7
sellorm: I initially tried without vim and wget, but I was getting a missing dependency error when the resulting package was installed that I struggled to diagnose. I ended up putting them back in for this reason. I'll do some more investigation on that, I'm guessing that one of them pulls in another dependency that we need here, I'm just not sure what it is.

builder/package.centos-7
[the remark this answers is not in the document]
glin: Aw ok, but that makes sense. But at least this is a whole lot better than not having an installer :)

builder/package.centos-7
glin: Oh interesting, I can take a look as well

glin: I tried removing vim/wget and got this error from `sessionInfo()` on R 3.5. Is it the same thing you saw?
```r
> sessionInfo()
Error in system(paste(which, shQuote(names[i])), intern = TRUE, ignore.stderr = TRUE) : 
  error in running command
```

The error message isn't great, but it looks like it's coming from `Sys.which()` not finding the `which` command. I was able to get it working after doing a `yum install which`
```r
> traceback()
5: system(paste(which, shQuote(names[i])), intern = TRUE, ignore.stderr = TRUE)
4: withCallingHandlers(expr, warning = function(w) invokeRestart("muffleWarning"))
3: suppressWarnings(system(paste(which, shQuote(names[i])), intern = TRUE, 
       ignore.stderr = TRUE))
2: Sys.which("uname")
1: sessionInfo()
```

sellorm: Yeah, that's the one - it should be fixed now.

I've removed sudo, vim and wget and added which

builder/package.centos-7
sellorm: Post install script added

glin: Nice, the latest changes look good to me. I only noticed one minor issue while trying to uninstall. There were a bunch of empty directories left over, I think because the rpm doesn't know about `libRblas.so.keep`

Is there a way to make the rpm aware of that file? I don't see an obvious way since it's only created after installing.. If not, a workaround could be to create libRblas.so.keep before the packaging step.

```sh
[root@6911e123180b /]# yum -y remove R-3.5.3
Loaded plugins: fastestmirror, ovl
Resolving Dependencies
--> Running transaction check
---> Package R-3.5.3.x86_64 0:1-1 will be erased
--> Finished Dependency Resolution

Dependencies Resolved

===============================================================================================================================================================================
 Package                                   Arch                                     Version                                  Repository                                   Size
===============================================================================================================================================================================
Removing:
 R-3.5.3                                   x86_64                                   1-1                                      installed                                    95 M

Transaction Summary
===============================================================================================================================================================================
Remove  1 Package

Installed size: 95 M
Downloading packages:
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Erasing    : R-3.5.3-1-1.x86_64                                                                                                                                          1/1 
  Verifying  : R-3.5.3-1-1.x86_64                                                                                                                                          1/1 

Removed:
  R-3.5.3.x86_64 0:1-1                                                                                                                                                         

Complete!

[root@6911e123180b /]# ls /opt/R/3.5.3/    
bin  lib  share

[root@6911e123180b /]# ls /opt/R/3.5.3/lib/R/lib     
libRblas.so.keep
```

sellorm: it might be cleanest to use a --before-remove script to remove it, or perhaps not keep it in the first place?

glin: Either of those sounds fine to me. I doubt most users would want to swap the default BLAS back in, but it'd still be nice to keep if there's an easy way to remove it.

sellorm: I created a 'before-remove' script to remove /opt/R/3.5.3/lib/R/lib/libRblas.so.keep, but the RPM system still knows that something in there has been changed and refuses to remove the /opt/R/3.5.3 directory tree.

In the end I've used an 'after-remove' script that deletes everything from /opt/R/R_VERSION down.

glin: Oh weird, maybe it's because we modified the original `libRblas.so` and would've had leftover directories anyway. This seems like the best solution then.

PR #23, opened 2019-08-08
Tests for R binaries and packages

glin:
These are tests to help verify that the R binaries are configured/compiled properly, and that the R packages work.

Specifically, this installs the R deb/rpm in a base OS container, runs some R tests, and uninstalls the deb/rpm. The R tests were copied from https://github.com/rstudio/r-docker, but I only pulled in the tests relevant to r-builds (e.g. BLAS is configured properly, no missing runtime libraries, help pages work).

To test locally, first build the R packages:
```sh
# Build R 3.5.3 on each OS
R_VERSION=3.5.3 make docker-build-r

# Or build on a specific OS
cd builder && R_VERSION=3.5.3 docker-compose up ubuntu-1604

# Note: this outputs packages to builder/integration/tmp. To rebuild,
# clear out any existing packages in that directory
```

And then run the tests:
```sh
# Test R 3.5.3 on each OS
R_VERSION=3.5.3 make docker-test-r

# Or test on a specific OS
cd test && R_VERSION=3.5.3 docker-compose up ubuntu-1604
```

PR #24, opened 2019-08-08
Run tests in Jenkins

glin:
This runs the tests from #23 in Jenkins, so we only push builder images to ECR if they produce working R binaries/packages. Since the r-builds matrix is huge, I only added tests for the latest patch of each minor version since 3.1.3. One of each minor version should still give pretty good coverage.

A big downside is that the tests add a lot of time to the Jenkins job -- about 40 mins extra right now. It might be possible to improve this with some Jenkinsfile tweaks, or we could pare down the build matrix. I did try the docker-4x agent originally, which built one R version in ~10 mins, but I think there's only one of those available at a time. Or, maybe the extra wait is fine since it still beats manual testing, and since r-builds doesn't change often anyway.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,52 @@ pipeline {
         sh 'make docker-build'
       }
     }
+    stage('test images') {
+      parallel {
+        stage('R 3.1') {
+          agent { label 'docker' }
+          steps {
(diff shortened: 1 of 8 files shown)


joncfoo:
:tada: thanks for putting this in!

glin:
Closing - I've been testing out a better approach using GitHub Actions and will open a new PR.

PR #30, opened 2019-08-23
Set default HTTP user agent

### LINKED ISSUE #15: Set User Agent Header
Beginning in R versions > 3.6, the user agent header is no longer set by default. We should include a user agent header that specifies the R version, and potentially the distribution id; this would alleviate the need for admins and our products to get the UA header correct, see:

https://github.com/rstudio/connect/issues/14583
https://github.com/rstudio/rstudio/pull/4889
https://community.rstudio.com/t/url-encoding-or-user-agent-encoding/35532

glin:
Resolves #15 

This configures the default `HTTPUserAgent` for R 3.6.0 and later in Rprofile.site. On R 3.6.1, it would be:
```
R/3.6.1 R (3.6.1 x86_64-pc-linux-gnu x86_64 linux-gnu)
```

Testing that the user agent is surfaced in R 3.6.0+:
```sh
root@637f112ecf62:/# /opt/R/3.6.0/bin/R

R version 3.6.0 (2019-04-26) -- "Planting of a Tree"
Copyright (C) 2019 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

...

> getOption("HTTPUserAgent")
# [1] "R/3.6.0 R (3.6.0 x86_64-pc-linux-gnu x86_64 linux-gnu)"

> options(internet.info = 1)
> install.packages("R6", repos = "https://cran.rstudio.com/")
# ...
# > GET /src/contrib/PACKAGES.rds HTTP/1.1
# Host: cran.rstudio.com
# User-Agent: R/3.6.0 R (3.6.0 x86_64-pc-linux-gnu x86_64 linux-gnu)
# Accept: */*
# ...

root@637f112ecf62:/# cat /opt/R/3.6.0/lib/R/etc/Rprofile.site 
# Set default HTTP user agent
options(HTTPUserAgent = sprintf("R/%s R (%s)", getRversion(), paste(getRversion(), R.version$platform, R.version$arch, R.version$os)))
```

---

The user agent format could easily change if needed. We also considered adding the OS identifier or "RStudio R", but thought it'd be better to keep the same amount of information included in the default user agent.

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -95,10 +95,12 @@ compile_r() {
   make install
 
   # Preserve the default HTTP user agent for R 3.6.0 and later
-  cat <<'EOF' >> /opt/R/${1}/lib/R/etc/Rprofile.site
+  if _version_is_greater_than ${1} 3.6; then
+    cat <<'EOF' >> /opt/R/${1}/lib/R/etc/Rprofile.site
 # Set default HTTP user agent
 options(HTTPUserAgent = sprintf("R/%s R (%s)", getRversion(), paste(getRversion(), R.version$platform, R.version$arch, R.version$os)))
 EOF
+  fi
 }
 
 # check for packager script
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -93,6 +93,12 @@ compile_r() {
   make clean
   make
   make install
+
+  # Preserve the default HTTP user agent for R 3.6.0 and later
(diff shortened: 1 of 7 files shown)


glin:
I've got some changes to work around the IDE issue and add `/etc/os-release` information, but I'll put those in a follow-up PR. The RSPM release w/ binaries is just about ready to go, so we'd like to make sure we have R 3.6 support before then.

PR #31, opened 2019-08-28
Quick Install Script

jonyoder:
This adds a quick install script to help ease installation. Before we merge this, we should:

- [x] Make a URL shortcut named `https://rstd.io/r-install` that points to the new `install.sh` script.

To test, since `https://rstd.io/r-install` points to a location (`master`) that doesn't yet exist, you can use this one-liner:

```sh
bash -c "$(curl -L https://raw.githubusercontent.com/rstudio/r-builds/jon/quick-install/install.sh)"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,16 @@ R binaries are built for the following Linux operating systems:
 
 ## Installation
 
+### Quick Installer
+
+To use our quick install script to install R, simply run the following
+command. To use the quick installer, you must have root or sudo privileges,
+and `curl` must be installed.
+
+```sh
+bash -c "$(curl -L https://rstd.io/r-install)"
+```
+
 ### Specify R version
 
 Define the version of R that you want to install. Available versions
@@ -34,7 +44,7 @@ of R can be found here: https://cdn.rstudio.com/r/versions.json
 R_VERSION=3.5.3
 ```
 
-### Download and install R
+### Download and install R Manually
 
 #### Ubuntu/Debian Linux
 
--- a/README.md
+++ b/README.md
@@ -24,9 +24,7 @@ R binaries are built for the following Linux operating systems:
 - openSUSE 42.3, 15.0
 - SUSE Linux Enterprise 12, 15
 
-## Installation
-
-### Quick Installer
+## Quick Installation
 
 To use our quick install script to install R, simply run the following
 command. To use the quick installer, you must have root or sudo privileges,
@@ -36,6 +34,8 @@ and `curl` must be installed.
 bash -c "$(curl -L https://rstd.io/r-install)"
 ```
 
+## Manual Installation
+
 ### Specify R version
 
 Define the version of R that you want to install. Available versions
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,405 @@
...
+  # Matches the JSON line that contains the r versions
+  grep r_versions | \
+  # Gets the value of the `r_version` property (e.g., "[ 3.0.0, 3.0.3, ... ]")
+  cut -f2 -d ":" | \
+  # Removes the opening and closing brackets of the array
+  cut -f2 -d "[" | cut -f1 -d "]" | \
+  # Removes the quotes and commas from the values
+  sed -e 's/\"//g' | sed -e 's/\,//g' | \
+  # Reverses the order of the list
+  ( while read -d ' ' f;do g="$f${g+ }$g" ;done;echo "$g" ))
```
glin: R 3.6.1 isn't showing up in the list of available versions, and I think it has to do with this line. For some reason, the read command is excluding the last item in the array?

README.md
```diff
@@ -34,7 +44,7 @@ of R can be found here: https://cdn.rstudio.com/r/versions.json
 R_VERSION=3.5.3
 ```
 
-### Download and install R
+### Download and install R Manually
```
glin: The "Specify R version" section above this is also part of the manual install process. What if we wrote something like this at the bottom of the Quick Installer section: "Or follow the steps below to install R manually". Or we could split the Installation section into Quick Installer and Manual Installation subsections.

install.sh
```diff
@@ -0,0 +1,405 @@
...
+install_pre () {
+  os=$1
+  ver=$2
+
+  case $os in
+    "RedHat" | "CentOS")
+      install_epel "${ver}"
+      ;;
+    "SUSE")
+      install_sci
```
glin: This will install the SLES 12 Science repo on openSUSE 42 as well, which isn't necessary. Is there a way to distinguish between openSUSE/SLES?

jonyoder:
@glin I merged @ricardofandrade's improvements into this branch and made a few changes based on your feedback.

install.sh
```diff
@@ -0,0 +1,443 @@
...
+  install "${installer_type}" "${installer_file_name}" "${os}" "${os_ver}"
+}
+
+do_show_usage() {
+  echo "r-builds quick install version ${THIS_VERSION}"
+  echo "Usage: `basename $0` [-i|-r|-v|-h|install|rversions|version|help]"
+  echo "Where:"
+  echo "'-i' or 'install' [version] (default) install the given version or list R versions available for quick install and prompt for one if none is provided"
+  echo "'-r' or 'rversions' list the R versions available for quick install, one per line"
+  echo "'-v' or 'versions' shows the version of this command"
```
ricardofandrade: Small mistake:
```suggestion
  echo "'-v' or 'version' shows the version of this command"
```

install.sh
```diff
@@ -0,0 +1,443 @@
+#!/bin/bash
+THIS_VERSION="1.0"
```
ricardofandrade:
```suggestion
THIS_VERSION="1.0.0"
```
We may want to go with a full semantic versioning here.

PR #39, opened 2019-10-03
Fix incompatibility with default gcc on CentOS/RHEL 6

glin:
@aronatkins discovered that when you install R on a CentOS 6 system with the default gcc, some packages like `digest` fail to install:
```r
> install.packages("digest")

* installing *source* package 'digest' ...
** package 'digest' successfully unpacked and MD5 sums checked
** using staged installation
** libs
g++ -std=gnu++11 -I"/opt/R/3.6.1/lib/R/include" -DNDEBUG   -I/usr/local/include  -fpic  -g -O2  -c SpookyV2.cpp -o SpookyV2.o
cc1plus: error: unrecognized command line option "-std=gnu++11"
```
But `digest` installs fine when using R from EPEL 6.

The problem is that we set the `-std=gnu++11` flag for packages that require C++11:
https://github.com/rstudio/r-builds/blob/7182392191dc455a2787370f25d6f499bfeef1d9/builder/package.centos-6#L23-L27

But the default gcc on CentOS/RHEL 6 (gcc 4.4.7) isn't new enough to support `-std=gnu++11`, only `-std=gnu++0x`: https://gcc.gnu.org/projects/cxx-status.html#cxx11. So any R packages that compile with C++11 (like `digest`) will immediately fail.

This switches the gcc flag to `-std=gnu++0x` for compatibility with the default gcc. I'm not real familiar with the differences between `-std=gnu++0x` and `-std=gnu++11`, but it sounds like they're equivalent -- `gnu++0x` is just deprecated in newer gcc versions: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-C_002b_002b11-code

### DIFF
--- a/builder/package.centos-6
+++ b/builder/package.centos-6
@@ -21,8 +21,9 @@ export GEM_HOME=/.gems
 export GEM_PATH=/.gems
 
 # Additional config for C++11 support in R 3.6.
-# R 3.6 no longer sets CXX11 flags without a full C++11 compiler present, but
+# R 3.6.0 no longer sets CXX11 flags without a full C++11 compiler present, but
 # the default gcc 4.4.7 still partially supports C++11 through -std=gnu++0x.
+# Use the same CXX11 flags as the defaults in R versions prior to 3.6.0.
 sed -i 's/^CXX11\ =\ .*/CXX11\ =\ g++/g' /opt/R/${R_VERSION}/lib/R/etc/Makeconf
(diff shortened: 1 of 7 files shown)


glin:
The flags were taken from the defaults on R versions prior to 3.6.0, so I think that was just to keep the script simpler and avoid having to do a version check. I've added an additional note about this in the script.

PR #46, opened 2019-10-31
shift deploy to separate job

stevenolen:
This moves the serverless deploy to a separate job so that different environments can be targeted.

Notably, the creation of new docker images is not part of the deploy stage, and is not based per-environment -- master branch builds will push new docker images.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,19 +23,13 @@ pipeline {
       }
     }
     stage('deploy') {
-      agent {
-        dockerfile {
-          dir 'jenkins'
-          label 'docker-4x'
-          additionalBuildArgs '--build-arg DOCKER_GID=$(stat -c %g /var/run/docker.sock) --build-arg JENKINS_UID=$(id -u jenkins) --build-arg JENKINS_GID=$(id -g jenkins)'
-          args '-v /var/run/docker.sock:/var/run/docker.sock --group-add docker'
-        }
-      }
+      agent none
       when {
         branch 'master'
       }
       steps {
-        sh 'make serverless-deploy.production'
+        build(job: 'hostedapps/deploy/r-builds', wait: true,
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -23,19 +23,13 @@ pipeline {
       }
     }
     stage('deploy') {
-      agent {
-        dockerfile {
-          dir 'jenkins'
-          label 'docker-4x'
```
jspiewak: really? any particular reason it was using `docker-4x`?

stevenolen: `docker-push` make target runs `docker-build` target first, so it's ok!

it uses docker-4x because the rest of the stages used it (the build stages build a pile of images with docker-compose in parallel, which allows them to complete in a reasonable amount of time). There was not real reason the deploy was using 4x though, so i removed it :)

deploy.Jenkinsfile
```diff
@@ -0,0 +1,27 @@
...
+    ansiColor('xterm')
+  }
+  parameters {
+    choice(name: 'ENVIRONMENT', choices: ['staging', 'production'],
+           description: 'The target environment to deploy to.')
+  }
+  stages {
+    stage('deploy') {
+      steps {
+        sh 'make serverless-deploy.${params.ENVIRONMENT}'
```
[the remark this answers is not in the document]
stevenolen: nope! thanks for catching 😄

stevenolen:
@glin -- it's possible, but not a pattern we really use on the hosted side of things!

stevenolen:
This has been deployed, and i test deploying to staging (via allowing the master branch to trigger the build) and deploying to production (triggered the job manually from https://buildmaster.rstudioservices.com/job/hostedapps/job/deploy/job/r-builds/)

@glin keep in mind for future merges to master that the behavior has changed -- production wont receive new changes to the serverless project that hit master without triggering the build with environment targeting `production`

PR #55, opened 2020-04-27
Add pcre2

jspiewak:
Which is actually newer than what is called libpcre3.

### DIFF
--- a/builder/package.ubuntu-1604
+++ b/builder/package.ubuntu-1604
@@ -33,6 +33,7 @@ fpm \
   -d libpango-1.0-0 \
   -d libpangocairo-1.0-0 \
   -d libpaper-utils \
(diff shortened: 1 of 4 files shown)


builder/package.ubuntu-1604
```diff
@@ -33,6 +33,7 @@ fpm \
   -d libpango-1.0-0 \
   -d libpangocairo-1.0-0 \
   -d libpaper-utils \
+  -d libpcre2 \
```
glin, line 36: This should be `libpcre2-8` (and ditto for Ubuntu 18). But otherwise, this looks good to me.

glin: oops: I meant `libpcre2-8-0`

PR #60, opened 2020-05-04
Including libopenblas fails to build on Ubuntu 20

joncfoo:
Builds on Ubuntu 20 were getting stuck on `byte compiling grDevices` as evidenced by build instances that had to be manually terminated after _days_ :disappointed:

Removing `libopenblas` as a build dependency allows the builds to complete as a workaround. FWICT, this change ends up using `libblas3` which gets pulled in from `apt-get build-dep r-base`

### DIFF
--- a/builder/package.ubuntu-2004
+++ b/builder/package.ubuntu-2004
@@ -21,6 +21,7 @@ fpm \
   -d gcc \
   -d gfortran \
   -d libbz2-1.0 \
+  -d libblas3 \
   -d libc6 \
   -d libcairo2 \
   -d libcurl4 \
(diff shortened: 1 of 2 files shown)


builder/package.ubuntu-2004
```diff
@@ -21,6 +21,7 @@ fpm \
   -d gcc \
   -d gfortran \
   -d libbz2-1.0 \
+  -d libblas3 \
```
glin, line 24: R also needs `liblapack3` here (libopenblas includes both BLAS and LAPACK by itself). Or the dev versions: `libblas-dev` and `liblapack-dev`. We've been including the `-dev` packages in the installers since a lot of R packages need them to install from source.

joncfoo: Updated to include `libblas-dev` and `liblapack-dev` as dependencies.

joncfoo:
I've filed https://github.com/rstudio/r-builds/issues/62 to note of the build hang issue that we noticed.

PR #65, opened 2020-06-02
Fix ffi version for older distros where ruby<2.3

joncfoo:
Use a specific `ffi` gem that is compatible with ruby < 2.3 for older distros.


joncfoo:
@glin this seems to fix the `ffi` issues.

PR #67, opened 2020-06-09
Pass through custom user-agent if specified

joncfoo:
The Connect setup-tool uses this script to download R binaries.  This change allows us to to appropriately count downloads that originated from the Connect setup-tool.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -352,6 +352,7 @@ do_download () {
     # Or, If curl is around, use that.
     elif [[ -z "${curl_rc}" ]]; then
         echo "Downloading ${url}..."
+        local header
         if [[ -n "${RS_USER_AGENT}" ]]; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -338,19 +338,28 @@ do_download () {
 
     if [[ -z "${wget_rc}" ]]; then
         echo "Downloading ${url}..."
+        local header
+        if [[ -n "${RS_USER_AGENT}" ]]; then
+          header="-d --header 'User-Agent: ${RS_USER_AGENT}'"
```
glin, line 343: Just double checking, did you want the `-d` debug flag in here? I noticed a similar flag wasn't added for curl.

joncfoo: Ah, thanks for that catch!  I've removed it.

jonyoder:
This is great!

PR #68, opened 2020-06-09
Fix header argument to curl and wget

joncfoo:
Fix header argument being passed incorrectly to curl/wget.  The script works again. Oops!

This change makes it so:
1. `wget` is passed the right argument `--header=string` in exactly that format. Otherwise it didn't work (on CentOS 6)
2. Split the arguments in the `header` variable, if any, to ensure argument splitting occurs when calling out to curl/wget.
    - w/o this the whole built-up string was being passed as a single argument which is not what we want (and didn't work)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -360,7 +360,7 @@ do_download () {
         if [[ "${RUN_UNATTENDED}" -ne "0" ]]; then
           curl -fsSL "${header[@]}" --output "${file_name}" "${url}"
         else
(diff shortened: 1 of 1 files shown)


joncfoo:
I'll have another look at this.  What do you think of setting a default User-Agent, e.g. `r-builds`?  It'll simplify some of the logic.

glin:
Agreed, that sounds like a great idea

PR #70, opened 2020-07-09
Update repository indexes before installing (fixes #66)

### LINKED ISSUE #66: Package caches should be updated before installation attempts within install script
We found this within the setup app as the server was executing the installation scripts for R. To resolve this, we had to run `apt update` ourselves, although the error we received did not make this obvious:
> Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/t/tzdata/tzdata_2019c-0ubuntu0.16.04_all.deb 404 Not Found [IP: 91.189.88.152 80]

Connect's Setup app executes this automatically.. we believe the r-builds installation script should do the same. 
- Code reference from Setup App: https://github.com/rstudio/connect-setup/blob/87bba60e1aa5835d742373867c3e321567f2ac87/scripts/install/install.sh#L158-L181

joncfoo:
Sometimes package repositories are out of date on a system and so installing R fails b/c the package manager has stale information about known dependencies.

This PR updates the system's repository indexes before installing R.

Fixes this issue: https://github.com/rstudio/r-builds/issues/66

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -248,6 +248,9 @@ install_deb () {
       yes="--n"
       yesapt="-y"
   fi
+  echo "Updating package indexes..."
+  ${SUDO} apt-get update -y
+  echo "Installing ${installer_name}..."
   ${SUDO} apt-get install ${yesapt} gdebi-core
   ${SUDO} gdebi ${yes} "${installer_name}"
 }
@@ -269,6 +272,9 @@ install_rpm () {
         echo "Must have sudo privileges to run yum"
         exit 1
       fi
+      echo "Updating package indexes..."
+      ${SUDO} yum check-update -y
+      echo "Installing ${installer_name}..."
       ${SUDO} yum install ${yes} "${installer_name}"
       ;;
     "LEAP12" | "LEAP15" | "SLES12" | "SLES15")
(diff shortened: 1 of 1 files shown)


slopp:
Perhaps this worry is overkill, but I'll throw it out there so at least we can be sure it has been considered.

As an admin, this type of update action hidden in an install script may feel surprising. For something like the setup tool where the script is run by non-Linux admins on a virgin server I can see the appeal in automating the `update` command.

But for other users of the `install-r.sh` script, I worry that an install script that runs update is not safely "minding its own business" and would surprisingly have a side affect that impacts other parts of the system

Why wouldn't we just modify the instructions for calling the script to suggest running the update? And then automate the update command in the setup tool?

joncfoo:
These _update_ commands do not update/upgrade any installed packages.  They only update the metadata used by the package managers so that when installing a `deb` or `rpm` the latest package information is known to `apt/yum/zypper`.

A user of this script can easily be in this state. e.g. a system that rarely has `apt-get update` run on it.

install.sh
```diff
@@ -248,6 +248,9 @@ install_deb () {
       yes="--n"
       yesapt="-y"
   fi
+  echo "Updating package indexes..."
+  ${SUDO} apt-get update -y
```
glin, line 252: I don't think `apt-get update` needs `-y` - at least, I've never seen a prompt from it

joncfoo: updated to remove `-y`

PR #73, opened 2020-08-31
Add builds for openSUSE 15.2

glin:
openSUSE 15.2 moved from ICU 60 to 65, so we'll need separate builds for openSUSE 15.0/15.1 and openSUSE 15.2. If you install the openSUSE 15.0 based package on 15.2, you'll get an error like this when running R:
```sh
$ /opt/R/3.6.3/bin/R
/opt/R/3.6.3/lib/R/bin/exec/R: error while loading shared libraries: libicuuc.so.60.2: cannot open shared object file: No such file or directory
```
The openSUSE 15.0 package may actually work on SLES 15 SP2 because it still has the `libicu60_2` runtime lib for compatibility, but it'll still be better to use an openSUSE 15.2 based package for SLES 15 SP2.

The Dockerfile and packaging scripts for 15.2 are the exact same as for 15.0. `libicu-devel` just pulls in a newer ICU now.

Also, I didn't update the quick install script here. That'll be easier to update and test after we get openSUSE 15.2 installers up.


glin:
Yup, the install script will need changes. I'll follow up with a PR for that after openSUSE 15.2 installers are up and tested.

Issue #82, opened 2020-12-17
R 3.6.3 fails to install on SLES 12.5


joncfoo:
<details>
<summary>Logs</summary>

```
 Downloading installer from https://cdn.rstudio.com/r/opensuse-42/pkgs/R-3.6.3-1-1.x86_64.rpm...

Downloading https://cdn.rstudio.com/r/opensuse-42/pkgs/R-3.6.3-1-1.x86_64.rpm...

User install from RPM installer R-3.6.3-1-1.x86_64.rpm...

Adding repository 'Backport builds of Python Modules (SLE_12_SP5)' [......done]

Repository 'Backport builds of Python Modules (SLE_12_SP5)' successfully added

URI : https://download.opensuse.org/repositories/devel:/languages:/python:/backports/SLE_12_SP5/

Enabled : Yes

GPG Check : Yes

Autorefresh : No

Priority : 99 (default priority)

Repository priorities are without effect. All enabled repositories share the same priority.

Updating package indexes...

Repository 'SLE-Module-Adv-Systems-Management12-Pool' is up to date.

Repository 'SLE-Module-Adv-Systems-Management12-Updates' is up to date.

Repository 'SLE-Module-Containers12-Pool' is up to date.

Repository 'SLE-Module-Containers12-Updates' is up to date.

Repository 'SLE-Module-HPC12-Pool' is up to date.

Repository 'SLE-Module-HPC12-Updates' is up to date.

Repository 'SLE-Module-Legacy12-Pool' is up to date.

Repository 'SLE-Module-Legacy12-Updates' is up to date.

Repository 'SLE-Module-Public-Cloud12-Pool' is up to date.

Repository 'SLE-Module-Public-Cloud12-Updates' is up to date.

Repository 'SLES12-SP5-Pool' is up to date.

Repository 'SLES12-SP5-Updates' is up to date.

Repository 'SLE-SDK12-SP5-Pool' is up to date.

Repository 'SLE-SDK12-SP5-Updates' is up to date.

Repository 'SLE-Module-Toolchain12-Pool' is up to date.

Repository 'SLE-Module-Toolchain12-Updates' is up to date.

Repository 'SLE-Module-Web-Scripting12-Pool' is up to date.

Repository 'SLE-Module-Web-Scripting12-Updates' is up to date.

Retrieving repository 'Backport builds of Python Modules (SLE_12_SP5)' metadata [..

New repository or package signing key received:

[...]

PR #83, opened 2020-12-30
Work around broken OpenBLAS dependency for SLES 12

glin:
For https://github.com/rstudio/r-builds/issues/82

On SLES 12, installing `openblas-devel` from the python-backports repo currently doesn't work because it depends on a newer `libopenblas_pthreads-devel` dependency that has not been published:
```sh
$ VERSION="SLE_$(grep "^VERSION=" /etc/os-release | sed -e 's/VERSION=//' -e 's/"//g' -e 's/-/_/')"
$ sudo zypper --gpg-auto-import-keys addrepo https://download.opensuse.org/repositories/devel:/languages:/python:/backports/$VERSION/devel:languages:python:backports.repo
$ sudo zypper install openblas-devel
...
Problem: nothing provides libopenblas_pthreads-devel = 0.3.12 needed by openblas-devel-0.3.12-2.5.x86_64
 Solution 1: do not install openblas-devel-0.3.12-2.5.x86_64
 Solution 2: break openblas-devel-0.3.12-2.5.x86_64 by ignoring some of its dependencies

Choose from above solutions by number or cancel [1/2/c] (c): 

# libopenblas_pthreads-devel is only available in version 0.3.10, not 0.3.12
$ sudo zypper search -s libopenblas_pthreads-devel
S | Name                       | Type    | Version    | Arch   | Repository                                    
--+----------------------------+---------+------------+--------+-----------------------------------------------
  | libopenblas_pthreads-devel | package | 0.3.10-6.3 | x86_64 | Backport builds of Python Modules (SLE_12_SP5)
```

To work around this, we can switch the dependency to `libopenblas_pthreads-devel`, which provides the same library but does install properly (version 0.3.10). The difference between the two packages is that `openblas-devel` provides `libopenblas.so`, while `libopenblas_pthreads-devel` provides `libopenblas_pthreads.so`. Both symlink to the same library:
```sh
$ rpm -ql openblas-devel
/usr/lib64/cmake
/usr/lib64/cmake/openblas
/usr/lib64/cmake/openblas/OpenBLASConfig.cmake
/usr/lib64/cmake/openblas/OpenBLASConfigVersion.cmake
/usr/lib64/libopenblas.so

$ rpm -ql libopenblas_pthreads-devel
/usr/lib64/libopenblas_pthreads.so
/usr/lib64/libopenblasp.so

$ readlink -f /usr/lib64/libopenblas.so
/usr/lib64/libopenblas_pthreads.so.0

$ readlink -f /usr/lib64/libopenblas_pthreads.so
/usr/lib64/libopenblas_pthreads.so.0
```

So users should be able to install this new opensuse-42 RPM over an old one without breaking existing packages that use BLAS.

### DIFF
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -29,6 +29,8 @@ RUN yum -y update \
     pango-devel \
     pcre-devel \
     pcre2-devel \
+    python3 \
+    python3-pip \
     readline-devel \
     rpm-build \
     ruby \
@@ -45,12 +47,8 @@ RUN yum -y update \
     zlib-devel \
     && yum clean all
 
-# Install pip for python
-RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
-    python get-pip.py
-
 # Install AWS CLI.
-RUN pip install awscli --upgrade --user && \
+RUN pip3 install awscli --upgrade --user && \
     ln -s /root/.local/bin/aws /usr/bin/aws
 
 RUN gem install ffi:1.12.2 fpm
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -51,7 +51,8 @@ RUN yum -y update \
 RUN pip3 install awscli --upgrade --user && \
     ln -s /root/.local/bin/aws /usr/bin/aws
 
-RUN gem install ffi:1.12.2 fpm
+# Pin fpm for compatibility with Ruby < 2.3
+RUN gem install ffi:1.12.2 fpm:1.11.0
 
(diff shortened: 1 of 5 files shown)


kevinhankens (approved):
I haven't tested, but if they are the same thing, then awesome!

glin:
FYI, I had to push some unrelated fixes to get this build working:
- pip dropped support for Python 2, so switch to Python 3 in the CentOS 7 image
- fpm released a new version that was incompatible with openSUSE 42 and CentOS 7, so pin fpm to the previous version for now

The other images will probably need fixes to handle issues with the new fpm release, but we can address that later.

PR #91, opened 2021-02-16
Pin fpm to 1.11.0

### LINKED ISSUE #86: Most builds fail with latest fpm 1.12.0 release
Continued from https://github.com/rstudio/r-builds/pull/83#issuecomment-767791975: 

The recent fpm 1.12.0 release adds a new runtime dependency on git, which causes most builds to fail since git isn't installed:

```sh
$ fpm
/usr/local/share/gems/gems/git-1.7.0/lib/git/lib.rb:989:in `command': git '-c' 'color.ui=false' version   2>&1:sh: git: command not found (Git::GitExecuteError)
```

This is fixed in the next fpm release after 1.12.0, so we can either wait for the new release or pin fpm. If the next fpm release fixes compatibility with Ruby < 2.3, we should also un-pin fpm on CentOS 7 and openSUSE 42 (https://github.com/rstudio/r-builds/pull/83/commits/3de0397ceeab6ff768313a366a4c3ec97785cfac).

glin:
Fixes #86

fpm 1.12.0 adds an unintended runtime dependency on git, which should be fixed in a future release of fpm (https://github.com/jordansissel/fpm/issues/1751). Pin fpm for now since installing git has its own side effects, such as installing `less` and changing the default pager on systems where `less` isn't available by default. It'll be more complicated to handle the pager issues (for example, https://github.com/rstudio/r-builds/pull/5).

### DIFF
--- a/builder/Dockerfile.centos-8
+++ b/builder/Dockerfile.centos-8
@@ -52,7 +52,8 @@ RUN dnf -y upgrade \
 RUN pip3 install awscli --upgrade --user \
     && ln -s /root/.local/bin/aws /usr/bin/aws
 
-RUN gem install fpm
+# Pin fpm to avoid git dependency in 1.12.0
+RUN gem install fpm:1.11.0
 
 RUN chmod 0777 /opt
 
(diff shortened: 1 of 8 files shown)


tylfin (approved):

kevinhankens (approved):

stevenolen:
I think it'd probably be more consistent/obvious to add a `Gemfile` with this version constraint, instead of installing a specific version in the dockerfiles

PR #95, opened 2021-10-21
build on opensuse15.3

dethmasque:
This PR adds building R binaries on OpenSUSE 15.3 for use with SLES 15sp3 and OpenSUSE 15.3.

### DIFF
--- a/builder/Dockerfile.opensuse-42
+++ b/builder/Dockerfile.opensuse-42
@@ -75,6 +75,15 @@ RUN gem install ffi:1.12.2 fpm:1.11.0 && \
 
 RUN chmod 0777 /opt
 
+# Work around issue with expired Let's Encrypt root certificate and OpenSSL 1.0.2,
+# which affects downloads from sourceforge.net (and possibly other sites).
+# On SLES 12, the issue can be resolved by updating ca-certificates-mozilla, but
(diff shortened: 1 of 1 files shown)


glin (commented):
Looks good, thanks for adding this!

There are just a couple other minor things to add outside of the build. Could we also:
1. Add openSUSE 15.3 / SLES 15 SP3 to the instructions in the README?
2. Add `opensuse-153` to the list of skipped OSs for an old R version bug: https://github.com/rstudio/r-builds/blob/d07acfe0cdb68fa99331ec1b12e92b169c05d447/handler.py#L128-L137
    This bug prevents a few R 3.3.x versions from building on newer OSs, including openSUSE 15.2, so I believe it'll affect openSUSE 15.3 as well.

    I realize this isn't obvious, so I've been thinking of inverting the check to exclude old OSs to make adding new platforms easier. But that's just an idea, and we don't have to do it now.

3. Update the quick install script to support openSUSE 15.3? Right now, it'll use the openSUSE 15.2 build if the version is 15.2 or higher: https://github.com/rstudio/r-builds/blob/d07acfe0cdb68fa99331ec1b12e92b169c05d447/install.sh#L179-L181

---

The CI failure seems to be a different issue with the openSUSE 42 Docker build. I took a look and found that expired Lets Encrypt root certificate issue from recent months. Trying out a fix at https://github.com/rstudio/r-builds/pull/96

PR #106, opened 2022-01-04
Add builds for Debian 11

glin:
Add builds for Debian 11. Based on the Debian 10 builds, but with a few differences to work around new issues in Debian 11:

### Include PCRE2 in R 3.x on Debian 11 for Pango support
In Debian 11, Pango now depends on PCRE2 (indirectly via one of its dependencies), so R 3.x will not be compiled with Pango support if we hide PCRE2 to prevent R 3.5 and 3.6 from taking an unnecessary dependency on PCRE2. For context, we hide PCRE2 from R 3.x builds because both PCRE1 and PCRE2 are installed in the build image, and R 3.5/3.6 will take an unused dependency on PCRE2 if its present (https://github.com/rstudio/r-builds/pull/63).

If R 3.x was compiled without Pango support, the configure script will print:

```
checking whether pkg-config knows about cairo and pango... no
```

The `bitmapType` option in R will default to "Xlib" instead of the usual "Cairo":

```r
> getOption("bitmapType")
"Xlib"
```

And internationalized text rendering will be off. For example, rendering a plot with Chinese text results in boxes:
```r
png("nopango.png", type = "cairo")
plot(1:3, main = "test 你好，世界")
dev.off()
```
plot with mis-rendered Chinese text without Pango support

### Fix compilation of R 3.6.1 and below due to changes in GCC 10

R <= 3.6.1 builds require additional compiler flags to work with GCC 10 and above:

- Add `-fcommon` to `CFLAGS` to work around an issue with the new default of `-fno-common` for C. This was fixed in R 3.6.2:
> * Source-code changes enable installation on platforms using gcc -fno-common (the expected default for gcc 10.x).
> https://cran.r-project.org/doc/manuals/r-release/NEWS.3.html

Without this flag, compilation fails with an error:

```sh
...
gcc -shared -fopenmp -L/usr/local/lib  -o libR.so CommandLineArgs.o Rdynload.o Renviron.o RNG.o agrep.o altclasses.o altrep.o apply.o arithmetic.o array.o attrib.o bind.o builtin.o character.o coerce.o colors.o complex.o connections.o context.o cum.o dcf.o datetime.o debug.o deparse.o devices.o dotcode.o dounzip.o dstruct.o duplicate.o edit.o engine.o envir.o errors.o eval.o format.o gevents.o gram.o gram-ex.o graphics.o grep.o identical.o inlined.o inspect.o internet.o iosupport.o lapack.o list.o localecharset.o logic.o main.o mapply.o match.o memory.o names.o objects.o options.o paste.o platform.o plot.o plot3d.o plotmath.o print.o printarray.o printvector.o printutils.o qsort.o radixsort.o random.o raw.o registration.o relop.o rlocale.o saveload.o scan.o seq.o serialize.o sort.o source.o split.o sprintf.o startup.o subassign.o subscript.o subset.o summary.o sysutils.o times.o unique.o util.o version.o g_alab_her.o g_cntrlify.o g_fontdb.o g_her_glyph.o xxxpr.o   `ls ../unix/*.o ../appl/*.o ../nmath/*.o` ../extra/tre/libtre.a    -lblas -lgfortran -lm -lquadmath   -lreadline  -lpcre2-8 -lpcre -llzma -lbz2 -lz -lrt -ldl -lm -licuuc -licui18n  
/usr/bin/ld: ../unix/system.o:/tmp/R-3.6.1/src/unix/../../src/include/Rinterface.h:157: multiple definition of `ptr_R_ProcessEvents'; ../unix/sys-unix.o:/tmp/R-3.6.1/src/unix/sys-unix.c:869: first defined here
collect2: error: ld returned 1 exit status
make[3]: *** [Makefile:177: libR.so] Error 1
make[3]: Leaving directory '/tmp/R-3.6.1/src/main'
```

- Add `-fallow-argument-mismatch` to `FFLAGS` to work around an issue with changes to argument mismatch checking in Fortran. This was fixed in R 3.6.2 but not mentioned in the NEWS for some reason. The only related reference I found was in the R Admin manual:
> * gfortran 10 by default gives a compilation error for the previously widespread practice of passing a Fortran array element where an array is expected, or a scalar instead of a length-one array. See https://gcc.gnu.org/gcc-10/porting_to.html.
> https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Using-Fortran

Without this flag, compilation fails with an error:

```sh
...
gfortran -fno-optimize-sibling-calls -fvisibility=hidden -fpic  -g -O2  -c lminfl.f -o lminfl.o
lminfl.f:117:57:

   95 |         call dqrsl(x, ldx, n, k, qraux, sigma, sigma, dummy,
      |                                               2          
......
  117 |                   call dqrsl(x, ldx, n, k, qraux, sigma, dummy, sigma,
      |                                                         1
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
make[5]: *** [../../../../etc/Makeconf:190: lminfl.o] Error 1
make[5]: *** Waiting for unfinished jobs....
make[5]: Leaving directory '/tmp/R-3.6.1/src/library/stats/src'
make[4]: *** [../../../share/make/basepkg.mk:140: mksrc] Error 1
make[4]: Leaving directory '/tmp/R-3.6.1/src/library/stats'
make[3]: *** [Makefile:30: all] Error 2
```

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -65,6 +65,22 @@ compile_r() {
     build_flag=''
   fi
 
+  # R 3.6.1 and below require additional compiler flags for GCC 10 and above
+  # (e.g., on Debian 11). Add -fcommon to CFLAGS to work around issues with new
+  # default of -fno-common for C (fixed in R 3.6.2). Add -fallow-argument-mismatch
+  # to FFLAGS to work around issues with changes to argument mismatch checking
+  # in Fortran (fixed in R 3.6.2, but not mentioned in NEWS).
+  # https://cran.r-project.org/doc/manuals/r-release/NEWS.3.html
+  # https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Using-Fortran
+  # https://gcc.gnu.org/gcc-10/porting_to.html
+  gcc_major_version=$(gcc -dumpversion | cut -d '.' -f 1)
+  if _version_is_less_than "${R_VERSION}" 3.6.2 && _version_is_greater_than "${gcc_major_version}" 9; then
+    extra_cflags='-fcommon'
+    extra_fflags='-fallow-argument-mismatch'
+    echo "Adding extra CFLAGS: ${extra_cflags}"
+    echo "Adding extra FFLAGS: ${extra_fflags}"
+  fi
+
   # Avoid a PCRE2 dependency for R 3.5 and 3.6. R 3.x uses PCRE1, but R 3.5+
   # will link against PCRE2 if present, although it is not actually used.
   # Since there's no way to disable this in the configure script, and we need
@@ -105,6 +121,8 @@ compile_r() {
   R_PRINTCMD=/usr/bin/lpr \
   R_UNZIPCMD=/usr/bin/unzip \
   R_ZIPCMD=/usr/bin/zip \
+  SHLIB_CFLAGS="${extra_cflags}" \
+  SHLIB_FFLAGS="${extra_fflags}" \
   ./configure \
     --prefix=/opt/R/${1} \
     ${CONFIGURE_OPTIONS} \
@@ -148,6 +166,10 @@ set_up_environment() {
 
 _version_is_greater_than() {
   test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1"
+}
+
+_version_is_less_than() {
+  test "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$2"
 }
 
 ###### RUN R COMPILE PROCEDURE ######
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -70,7 +70,12 @@ compile_r() {
   # Since there's no way to disable this in the configure script, and we need
   # PCRE2 for R 4.x, we hide PCRE2 from the configure script by temporarily
   # removing the pkg-config file and pcre2-config script.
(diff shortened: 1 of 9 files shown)


builder/build.sh
```diff
@@ -100,6 +121,8 @@ compile_r() {
   R_PRINTCMD=/usr/bin/lpr \
   R_UNZIPCMD=/usr/bin/unzip \
   R_ZIPCMD=/usr/bin/zip \
+  SHLIB_CFLAGS="${extra_cflags}" \
+  SHLIB_FFLAGS="${extra_fflags}" \
```
glin, line 125: I'm not 100% sure if `SHLIB_CFLAGS` and `SHLIB_FFLAGS` are the best environment variables to use, but they're the only way I've found to append CFLAGS and FFLAGS. Otherwise, you'll need to manually specify the CFLAGS and FFLAGS that R adds, which can differ across R versions.

gaborcsardi: @glin I think this also fixes the build errors with Ubuntu 22.04, right? Currently the builds for R 3.6.1 and below are missing for Ubuntu 22.04 AFAICT.

glin: Yup, I'll be sure to rebuild the failing R <= 3.6.1 builds for Ubuntu 22 when this gets deployed.

PR #112, opened 2022-02-10
Use Rocky Linux 8 in place of CentOS 8; platform updates

glin:
- Use Rocky Linux 8 in place of CentOS 8 to get R builds working for RHEL 8 again. Drop support for CentOS 8 since that's now EOL. We'll continue using `centos-8` for the RHEL 8 installer URLs to prevent breaking workflows for now.
- Platform updates: discontinue builds for the long-EOL Ubuntu 16 and openSUSE/SLES 15.1, and update the README for recently added Ubuntu 22 support.


builder/docker-compose.yml
```diff
@@ -78,37 +67,26 @@ services:
     image: r-builds:centos-7
     volumes:
       - ./integration/tmp:/tmp/output
-  # centos-8:
```
shepherdjerred, line 81: Can this be deleted instead of commented out?

shepherdjerred (approved):

PR #113, opened 2022-02-23
update builder (Docker and package files for better java support)

michaelmayer2:
This pull request will add better java support to our opinionated R builds. 

Main issue is that the build of R from sources will auto-detect the java version present on the system. This is great to start with. The problem however that it detects the JAVA_HOME and sets it to the physically existing path. Such a path not only contains the major version of Java but also any patch level or minor release versions as defined by the linux distro. In this pull request all the files in the directory builder are updated to set JAVA_HOME to a folder that is independent of any security patches or updates applied to the underlying linux distro. 

Accepting this pull request will fix any issue that leads to the urban myth that using Java in R is difficult. 

An article on solutions.rstudio.com is in preparation https://github.com/rstudio/solutions.rstudio.com/pull/210

### DIFF
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -11,8 +11,7 @@ RUN yum -y update \
     gcc-c++ \
     gcc-gfortran \
     gcc-objc \
-    java-1.8.0-openjdk-devel \
-    java-1.8.0-openjdk-headless \
+    java-11-openjdk-devel \
     libICE-devel \
     libSM-devel \
     libX11-devel \
@@ -66,6 +65,9 @@ ENV CONFIGURE_OPTIONS="\
     --with-tcl-config=/usr/lib64/tclConfig.sh \
     --with-tk-config=/usr/lib64/tkConfig.sh \
     --enable-prebuilt-html"
+
+# Make sure that patching the OS does not break Java
+ENV JAVA_HOME=/usr/lib/jvm/jre-11-openjdk
 
 # RHEL 7 doesn't have the inconsolata font, so override the defaults.
 ENV R_RD4PDF="times,hyper"
--- a/builder/Dockerfile.debian-10
+++ b/builder/Dockerfile.debian-10
@@ -17,6 +17,9 @@ RUN chmod 0777 /opt
 # Pin fpm to avoid git dependency in 1.12.0
 RUN gem install fpm:1.11.0
 
+# Make sure that patching the OS does not break Java
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/
+
 # Override the default pager used by R
 ENV PAGER /usr/bin/pager
 
--- a/builder/package.centos-7
+++ b/builder/package.centos-7
@@ -59,6 +59,7 @@ fpm \
   -d xz-devel \
   -d zip \
   -d zlib-devel \
+  -d java-11-openjdk-devel \ 
   /opt/R/${R_VERSION}
 
 shopt -s extglob
--- a/builder/Dockerfile.centos-8
+++ b/builder/Dockerfile.centos-8
@@ -12,8 +12,7 @@ RUN dnf -y upgrade \
     cairo-devel \
     gcc-c++ \
     gcc-gfortran \
-    java-1.8.0-openjdk-devel \
-    java-1.8.0-openjdk-headless \
+    java-11-openjdk-devel \
     libICE-devel \
     libSM-devel \
     libX11-devel \
@@ -67,6 +66,9 @@ ENV CONFIGURE_OPTIONS="\
     --with-tcl-config=/usr/lib64/tclConfig.sh \
     --with-tk-config=/usr/lib64/tkConfig.sh \
     --enable-prebuilt-html"
+
+# Make sure that patching the OS does not break Java
+ENV JAVA_HOME=/usr/lib/jvm/jre-11-openjdk
 
 # RHEL 8 doesn't have the inconsolata font, so override the defaults.
 ENV R_RD4PDF="times,hyper"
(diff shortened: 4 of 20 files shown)


builder/Dockerfile.debian-10
```diff
@@ -17,6 +17,9 @@ RUN chmod 0777 /opt
 # Pin fpm to avoid git dependency in 1.12.0
 RUN gem install fpm:1.11.0
 
+# Make sure that patching the OS does not break Java
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/
```
glin, line 21: On Ubuntu/Debian, the R builds here appear to already use this major-versioned location:
```sh
R CMD config JAVA_HOME
# /usr/lib/jvm/java-11-openjdk-amd64
```

Checking the default Ubuntu/Debian R packages, they appear to all use `/usr/lib/jvm/default-java` from `default-jdk` or `default-jre`:
```sh
R CMD config JAVA_HOME
# /usr/lib/jvm/default-java
```
Unfortunately it doesn't look like you can control the JRE location using the alternatives system in Ubuntu/Debian. `/usr/lib/jvm/default-java` is put there by `default-jdk`/`default-jre`, and not present if you install Java 11 or Java 8 specifically.
```sh
ls /usr/lib/jvm -la 
# lrwxrwxrwx 1 root root   25 Jul 17  2019 default-java -> java-1.11.0-openjdk-amd64
# lrwxrwxrwx 1 root root   21 Jul 20 22:37 java-1.11.0-openjdk-amd64 -> java-11-openjdk-amd64
# lrwxrwxrwx 1 root root   21 Jul 22 21:02 java-1.17.0-openjdk-amd64 -> java-17-openjdk-amd64
# drwxr-xr-x 9 root root 4096 Aug  8 23:34 java-11-openjdk-amd64
# drwxr-xr-x 9 root root 4096 Aug  8 23:34 java-17-openjdk-amd64
# drwxr-xr-x 2 root root 4096 Aug  8 23:34 openjdk-11
# drwxr-xr-x 2 root root 4096 Aug  8 23:34 openjdk-17
```
But still, maybe it's better to use `/usr/lib/jvm/default-java` anyway to match the distro default R and because it's common for users to install `default-jdk` (not sure, might be wrong about this)? At least for RSPM, `default-jdk` is what we've been using for instructions on installing Java.

builder/package.centos-7
```diff
@@ -59,6 +59,7 @@ fpm \
   -d xz-devel \
   -d zip \
   -d zlib-devel \
+  -d java-11-openjdk-devel \
```
glin, line 62: Since Java is often an optional dependency and somewhat heavy, could we leave it out of the DEB/RPM package dependencies? That would better support users who want to switch between Java 8, 11, 17, etc. as well. I think the default R packages from the various distros also leave Java as an optional dependency.

builder/Dockerfile.centos-7
```diff
@@ -67,6 +66,9 @@ ENV CONFIGURE_OPTIONS="\
     --with-tk-config=/usr/lib64/tkConfig.sh \
     --enable-prebuilt-html"
 
+# Make sure that patching the OS does not break Java
+ENV JAVA_HOME=/usr/lib/jvm/jre-11-openjdk
```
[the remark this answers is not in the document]
michaelmayer2: I would suggest we keep the version number encoded in `JAVA_HOME`, e.g. `/usr/lib/jvm/jre-11-openjdk`. As you point out, moving between java 11 and 17 seem to work but not when moving to 8 (not without a `javareconf`). 

The idea of this PR is to make Java integration a bit more seamless. I am not sure if we are creating more issues by moving to the extremely generic path compared to the issues that we are trying to solve.

builder/package.centos-7
michaelmayer2: We can leave java out but why do we then spend effort on configuring Java in the first place ?

glin: I'm not sure if there was a specific reason for configuring Java in the first place, and would guess that it was just borrowed from how Debian and Fedora packaged R. It is nice to be able to just install Java and have it work out of the box without `R CMD javareconf`. But I still do think Java is too heavy to add as a DEB/RPM dependency since these are billed as "minimal" R builds and Java is an optional dependency. It'll add a few hundred MBs to many consumers who don't necessarily need Java, like the RStudio R Docker images, the RStudio Pro Product images, the GitHub Actions for R, and others. And those on Java 8 or 17 will have an extra unused installation of Java.

glin:
@michaelmayer2 I went ahead and resolved the merge conflicts, just clearing out some of the removed EOL distros.

builder/package.centos-7
michaelmayer2: Ok, I definitely agree with your sentiment about the size of the R build if we go with the Java integration.

glin: Yeah, it's kind of both opinionated and minimal. We do some opinionated things that aren't completely minimal, like including OpenBLAS and the `-devel` packages of a few libraries that are required to install packages that link against the R shared library (which would only be included in `r-base-dev`, not `r-base` for example). A separate DEB/RPM that includes Java would be nice, but agreed that it'd complicate installation. Maybe a middle ground would be to provide users with a preconfigured Java where possible, like in the various Docker images. That article also looks great, and I'll try to reference it when any Java configuration questions come up.

Do you still want to explicitly set `JAVA_HOME` to at least make a first time Java installation more seamless? I've done that with the RHEL 9 R builds: https://github.com/rstudio/r-builds/pull/133/files#diff-7c6c557e074afff5ac851dbf63d384a5deaea84626a0f6945f41767510c08c39R77-R83. I think it's still better than letting R default it to something like `/usr/lib/jvm/java-11-openjdk-11.0.16.0.8-1.el9_0.x86_64`.

michaelmayer2: Yes, setting `JAVA_HOME` for the RHEL9 builds to the more generic folder I would strongly recommend. Also agree with the notion of extra DEB/RPM packages for Java where we even could be in the business of providing packages for any JDK offered, e.g. 1.8.0, 11, 17, ...

PR #116, opened 2022-03-30
Build the next version of R as well

gaborcsardi:
R-next is R-patched if there is no active release process, and R-alpha -> R-beta -> R-rc -> R-prerelease before a release. After the release it switches back to R-patched.

The name 'next' is not great, but 'latest' means "latest release" for many people, and `prerelease` is used by R core for the very last build before the actual release.

I tested Ubuntu 18.04 and 20.04 locally. Did not build on other OSes, but since we already build release and devel, it should be fine, I think. 

I also did not test that serverless stuff, obviously. If passing a two-element list to `version` works, then that should be fine as well.

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -46,11 +46,15 @@ fetch_r_source() {
   elif [ "${1}" = devel ]; then
     # Download the daily tarball of R devel
     wget -q https://stat.ethz.ch/R/daily/R-devel.tar.gz -O /tmp/R-devel.tar.gz
+  elif [ "${1}" = "next" ]; then
+    wget -q https://cran.r-project.org/src/base-prerelease/R-latest.tar.gz -O /tmp/R-next.tar.gz
   else
     wget -q "${CRAN}/src/base/R-`echo ${1}| awk 'BEGIN {FS="."} {print $1}'`/R-${1}.tar.gz" -O /tmp/R-${1}.tar.gz
   fi
   echo "Extracting R-${1}"
   tar xf /tmp/R-${1}.tar.gz -C /tmp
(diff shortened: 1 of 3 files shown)


gaborcsardi:
Hmmm, probably I should update the README as well.

builder/build.sh
```diff
@@ -46,11 +46,15 @@ fetch_r_source() {
...
     wget -q https://stat.ethz.ch/R/daily/R-devel.tar.gz -O /tmp/R-devel.tar.gz
+  elif [ "${1}" = "next" ]; then
+    wget -q https://cran.r-project.org/src/base-prerelease/R-latest.tar.gz -O /tmp/R-next.tar.gz
   else
     wget -q "${CRAN}/src/base/R-`echo ${1}| awk 'BEGIN {FS="."} {print $1}'`/R-${1}.tar.gz" -O /tmp/R-${1}.tar.gz
   fi
   echo "Extracting R-${1}"
   tar xf /tmp/R-${1}.tar.gz -C /tmp
+  dirname=`tar tzvf /tmp/R-next.tar.gz | head -1 | awk '{ print $NF }' | cut -d/ -f1`
+  mv /tmp/${dirname} /tmp/R-${1}
```
glin, line 57: Seems like this will cause issues if the R version isn't `next`, and the tarball name is different. Should we only do it when the version is `next`?

Besides this, the changes look great to me though.

gaborcsardi: Oh, right, good catch, silly me. Just fixed it. This time I checked the over versions as well.

glin (approved):

gaborcsardi:
Thanks for the review!

PR #127, opened 2022-07-26
Update instance types

### LINKED ISSUE #126: Ensure r-builds is using a more up to date image in it's launch template
This could possibly be as simple as redeploying it.  We should add an `r-builds-deploy-role` much like the `python-builds-deploy-role` and `revdepcheck-cloud-deploy-role`. This should be accomplished after https://github.com/rstudio/rstudio-pulumi/issues/37

Then we can add a jenkins job to deploy the project, just like in python-builds

jforest:
More modern and less expensive instance types, update the bid percentage to match python-builds as well.

Changing this will update the launch template, which will pull in a newer ami instead of the deprecated `ami-0d09143c6fc181fe3`

closes #126

### DIFF
--- a/serverless-resources.yml
+++ b/serverless-resources.yml
@@ -82,7 +82,7 @@ rBuildsEcsLaunchTemplate:
       BlockDeviceMappings:
         - DeviceName: /dev/xvdcz
           Ebs:
-            VolumeType: gp2
+            VolumeType: gp3
             VolumeSize: 64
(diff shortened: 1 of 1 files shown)


jspiewak:
Could you also switch to gp3 volumes please?

jforest:
> Could you also switch to gp3 volumes please?

Done!

PR #130, opened 2022-08-12
Add tests for R binaries and packages

glin:
Supersedes https://github.com/rstudio/r-builds/pull/23. This is almost the same PR but with updated platforms and a few improvements.

---

> These are tests to help verify that the R binaries are configured/compiled properly, and that the R packages work.
>
> 
> Specifically, this installs the R deb/rpm in a base OS container, runs some R tests, and uninstalls the deb/rpm. The R tests were copied from https://github.com/rstudio/r-docker, but I only pulled in the tests relevant to r-builds (e.g. BLAS is configured properly, no missing runtime libraries, help pages work).

Unlike https://github.com/rstudio/r-builds/pull/23 which had custom scripts to install the DEB/RPMs, this PR now uses the quick install script to install local builds of R, which both makes the tests easier to maintain and helps test the quick install script. I found a couple small bugs with the quick install script in the process and also fixed those here.

If there's no local build of R, the quick install script will download R from the CDN, so you can also run the tests on published builds without having to build R locally.

> To test the R builds locally, you can use the `build-r-$PLATFORM` and `test-r-$PLATFORM`
> targets to build R and run the tests. The tests use the quick install script to install R,
> using a locally built R if present, or otherwise a build from the CDN.
> 
> ```bash
> # Build R 4.1.3 for Ubuntu 22
> R_VERSION=4.1.3 make build-r-ubuntu-2204
> 
> # Test R 4.1.3 for Ubuntu 22
> R_VERSION=4.1.3 make test-r-ubuntu-2204
> ```

Unlike https://github.com/rstudio/r-builds/pull/23, the Make targets here have the platform built into the target name so you can autocomplete on `make build-r-[TAB]`

> Alternatively, you can build an image using the `docker-build-$PLATFORM` target, launch a bash session within a container using the `bash-$PLATFORM` target, and interactively run the build script:
> 
> ```bash
> # Build the image for Ubuntu 22
> make docker-build-ubuntu-2204
> 
> # Launch a bash session for Ubuntu 22
> make bash-ubuntu-2204
> 
> # Build R 4.1.3
> R_VERSION=4.1.3 ./build.sh
> 
> # Build R devel with parallel execution to speed up the build
> MAKEFLAGS=-j4 R_VERSION=devel ./build.sh
> 
> # Build a prerelease version of R (e.g., alpha or beta)
> R_VERSION=rc R_TARBALL_URL=https://cran.r-project.org/src/base-prerelease/R-latest.tar.gz ./build.sh
> ```

These are new convenience Make targets for development, also with autocomplete for the platform ID.


jonyoder (approved):
This looks great to me!

jonyoder:
Is it correct that this PR is configured to merge into the `suse154` branch?

Makefile
```diff
@@ -39,6 +39,26 @@ rebuild-all: deps fetch-serverless-custom-file
...
+build-r-$(platform):
+	@cd builder && R_VERSION=$(R_VERSION) docker-compose up $(platform)
+
+test-r-$(platform):
+	@cd test && R_VERSION=$(R_VERSION) docker-compose up $(platform)
+
+bash-$(platform):
+	docker run -it --rm --entrypoint /bin/bash -v $(CURDIR):/r-builds r-builds:$(platform)
+
+.PHONY: docker-build-$(platform) build-r-$(platform) test-r-$(platform) bash-$(platform)
```
shepherdjerred, line 55: This would be better using static pattern rules: https://www.gnu.org/software/make/manual/html_node/Static-Usage.html#Static-Usage

This would allow you to get rid of that `for` loop.

i.e.:
```make
docker-build-$(PLATFORMS):
	@cd builder && docker-compose build $(platform)
```

test/docker-compose.yml
```diff
@@ -0,0 +1,83 @@
+version: '2.0'
```
shepherdjerred, line 1: Any reason you chose `2.0` and not the latest (`3.8`)?

test/docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+
```
shepherdjerred, line 2: It'd be nice if we could templatize & generate this file since it's repetitive. If you're up for it [gomplate](https://docs.gomplate.ca/) would be an easy way to do it, but I'm also fine with you merging this as-is.

test/test-apt.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
shepherdjerred, line 1: These tests have a very similar structure. Could we reuse some of the code so there's less duplication?

shepherdjerred (approved):
Looks good to me overall. I had some suggestions on potential ways to improve the code, but I'm fine with you merging this as-is.

test/docker-compose.yml
glin: Good question, I copy pasted this file from the [builder's docker-compose.yml](https://github.com/rstudio/r-builds/blob/master/builder/docker-compose.yml) and didn't think about it. Looking up the version field now, it appears to have been deprecated and made optional in docker-compose.yml, so I'll try removing it.

https://docs.docker.com/compose/compose-file/#compose-file
> The Compose file is a [YAML](http://yaml.org/) file defining [version](https://docs.docker.com/compose/compose-file/#version-top-level-element) (DEPRECATED),

https://docs.docker.com/compose/
> A docker-compose.yml looks like this:
> ```yaml
> version: "3.9"  # optional since v1.27.0
> ```

test/docker-compose.yml
glin: There are a few things that are hard to predict and could make template generation tricky, like the Docker image name and whether to use apt/yum/zypper. There is some repetitive copy-pasta here (and also in other files in this repo), but I think it's okay for now since it only takes a few seconds to delete or add a new section.

test/test-apt.sh
glin: Most of the test code is shared in the `test-r.sh` and `test.R` scripts. These test-apt/yum/zypper scripts are more like bootstrap scripts that differ for the most part (apt vs. yum vs. zypper), and should rarely have to change, so I think it's ok for now.

glin (commented):
> Is it correct that this PR is configured to merge into the suse154 branch?
Yup, I did RHEL 9 and SUSE 15.4 support first, but wanted to get the test PRs up first to make the other PRs easier to review. I won't merge this until RHEL 9 / SUSE 15.4 support are in.

glin:
I figured it made more sense to merge the tests in first before RHEL 9 and SUSE 15.4, so just rebased with those platforms removed.

PR #131, opened 2022-08-12
Add GitHub Actions workflow to test R builds

glin:
Supersedes https://github.com/rstudio/r-builds/pull/24. This takes a different approach that replaces Jenkins with GitHub Actions, which is more open (external contributors can see logs now), capable of running large build matrices, and has a few built-in features that were really useful here.

---

Currently, it's hard to maintain and add new R builds because the tests are almost completely run manually. The Jenkins job only builds the Docker images, and R has to be built and tested manually across a large matrix of platforms and R versions. To make maintenance and contribution easier, this adds a GitHub Actions workflow that runs the tests from #130 so you can spot check the R builds before opening a PR. Since the build matrix is huge:
- **Tests only run on the last 5 minor R versions** by default (currently 4.2.1, 4.1.3, 4.0.5, 3.6.3, 3.5.3), same as the Tidyverse and RSPM supported versions. The versions are determined automatically from https://cdn.rstudio.com/r/versions.json, so there's no maintenance to update the R versions. The tests will take ~20-50 mins to run like this (depending on whether the build images are cached) which is pretty long, but saves you from having to run the same tests manually. This can be adjusted if the the tests start taking too long or hit the GHA limits.
- **Tests only run on changes to files in `builder/`, `test/`, or the `Makefile`.** You should be able to update other parts of the repo like the AWS infra without having to wait on unnecessary tests.
  - Tests can also be force-skipped using GHA's built-in feature of adding `[skip ci]` to commit messages
- **Tests can be triggered manually in GitHub Actions with custom parameters**. When adding a new platform, you can go to the Actions page and kick off a test with just a single platform or the full list of R versions down to R 3.1.

Manually trigger a test workflow in GitHub Actions

~~For future improvements, we could possibly cut time by pushing/pulling images to a public Docker Hub repo, but I didn't do that here.~~ I ended up adding Docker layer caching using Docker Buildx's local caching feature and GHA's built-in cache: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#local-cache. This cut the run time from 1.5 hours to ~20-50 mins, and doesn't use Docker Hub or any Docker registry.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,15 +43,13 @@ jobs:
         r_versions=$(python test/get_r_versions.py ${{ github.event.inputs.r_versions }})
         echo "::set-output name=r_versions::$r_versions"
 
-  test:
+  docker-images:
     needs: setup-matrix
     strategy:
       matrix:
         platform: ${{ fromJson(needs.setup-matrix.outputs.platforms) }}
-    env:
-      R_VERSIONS: ${{ join(fromJson(needs.setup-matrix.outputs.r_versions), ' ') }}
     runs-on: ubuntu-latest
-    name: ${{ matrix.platform }} (R ${{ join(fromJson(needs.setup-matrix.outputs.r_versions), ', ') }})
+    name: Docker image (${{ matrix.platform }})
     steps:
       - uses: actions/checkout@v3
 
@@ -75,23 +73,13 @@ jobs:
       # Use docker buildx instead of docker-compose here because cache exporting
       # does not seem to work as of docker-compose v2.6.0 and buildx v0.8.2, even
       # though it works with buildx individually.
-      - name: Build R
+      - name: Build image
         run: |
-          for version in ${{ env.R_VERSIONS }}; do
-            docker buildx build -t r-builds:${{ matrix.platform }} \
-              --file builder/Dockerfile.${{ matrix.platform }} \
-              --cache-from "type=local,src=/tmp/.buildx-cache" \
-              --cache-to "type=local,dest=/tmp/.buildx-cache-new,mode=max" \
-              --load \
-              builder
-            R_VERSION=$version make build-r-${{ matrix.platform }}
-          done
-
-      - name: Test R
-        run: |
-          for version in ${{ env.R_VERSIONS }}; do
-            R_VERSION=$version make test-r-${{ matrix.platform }}
-          done
+          docker buildx build -t r-builds:${{ matrix.platform }} \
+            --file builder/Dockerfile.${{ matrix.platform }} \
+            --cache-from "type=local,src=/tmp/.buildx-cache" \
+            --cache-to "type=local,dest=/tmp/.buildx-cache-new,mode=max" \
+            builder
 
       # Temporary workaround for unbounded GHA cache growth with the local cache mode.
       # https://github.com/docker/build-push-action/issues/252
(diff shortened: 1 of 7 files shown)


glin:
Hmm, I don't see the "Run workflow" button on this repo at https://github.com/rstudio/r-builds/actions/workflows/test.yml, but I was able to use it on a personal fork of the repo (where the screenshot comes from) :thinking:. Maybe a permissions thing. Does anyone else see the button?

shepherdjerred:
> Hmm, I don't see the "Run workflow" button on this repo at https://github.com/rstudio/r-builds/actions/workflows/test.yml, but I was able to use it on a personal fork of the repo (where the screenshot comes from) 🤔. Maybe a permissions thing. Does anyone else see the button?

I believe the button will only appear once this PR is merged in.

glin:
Ohh, d'oh, that makes total sense, good point.

shepherdjerred:
Is there a reason you chose not to add the R version as part of the build matrix? That could potentially lead to a 5x speedup.

shepherdjerred (approved):
This looks good to me. I left a comment above about adding the R version to the matrix.

Is there a particular reason you chose to use python over bash for the scripts? I personally prefer Python for small scripts, but I know others feel differently.

jonyoder (approved):
This looks reasonable (and good) to me. I'm not very familiar with GHA configurations or with Python, so I didn't have a lot of feedback.

glin:
> Is there a reason you chose not to add the R version as part of the build matrix? That could potentially lead to a 5x speedup.

Each job has to build a fresh image, so it's just saving GHA some $$ by sharing the same image build. But hmm, I wonder if it's possible to save the image as an artifact to be reused by other jobs, so at least the build-R step could be parallelized. Maybe possible looking at https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md, or we could just go ahead and push/pull to Docker Hub or GitHub's Container registry. I'll look into it today.

> Is there a particular reason you chose to use python over bash for the scripts? I personally prefer Python for small scripts like this, but I know others feel differently.

I started out with bash but it got really gnarly :grin:. This is what I had to extract the latest minor R versions from the versions.json file:
```sh
# Get all R versions 3.1 and above
R_VERSIONS=$(curl -s https://cdn.rstudio.com/r/versions.json |
  # Matches the JSON line that contains the r versions
  grep r_versions |
  # Gets the value of the `r_version` property (e.g., "[ 3.0.0, 3.0.3, ... ]")
  cut -f2 -d ":" |
  # Removes the opening and closing brackets of the array
  cut -f2 -d "[" | cut -f1 -d "]" |
  # Removes the quotes and commas from the values
  sed -e 's/\"//g' | sed -e 's/\,//g' |
  # Convert to newlines and sort in ascending order
  tr ' ' '\n' | sort |
  # Exclude devel/next versions
  grep '^[0-9]' |
  # Exclude R 3.0
  grep -v '^3.0')

# Get minor versions. Sort is needed because awk arrays don't have order.
echo "$R_VERSIONS" |
  awk '{ split($1,v,"."); minor_vers[v[1],v[2]]=$1 } END { for (i in minor_vers) print minor_vers[i] }' |
  sort
```

jq wouldn't have made it much better so I just converted it to Python. And Python because this project has historically used Python elsewhere, e.g. in handler.py.

shepherdjerred:
> Each job has to build a fresh image, so it's just saving GHA some $$ by sharing the same image build. But hmm, I wonder if it's possible to save the image as an artifact to be reused by other jobs, so at least the build-R step could be parallelized. Maybe possible looking at https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md, or we could just go ahead and push/pull to Docker Hub or GitHub's Container registry. I'll look into it today.

Oh that might be more trouble than it's worth. If we're not too concerned about performance/build times then it might be better to leave as-is and avoid the woes of caching with GHA. It's up to you!

glin:
@shepherdjerred I tried out local Docker image caching + parallelized R version builds and think it's worth the extra effort, although a little rough due to relying on fairly new features. It's using the local layer caching feature in Docker Buildx: https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#local-cache. There were some gotchas to work around, but Docker layer caching alone cut the build time by ~20 mins (fully cached), from 1h 30m to 1h 10m. And parallelizing the R version builds further cut the cached test time down to 50m, which is still pretty huge.

Restoring from the GHA cache is really fast, probably much faster than pulling an image from Docker Hub:
```sh
Run actions/cache@v2
Received 171966464 of 509757675 (33.7%), 163.8 MBs/sec
Received 381681664 of 509757675 (74.9%), 181.9 MBs/sec
Received 509757675 of 509757675 (100.0%), 157.5 MBs/sec
Cache Size: ~486 MB (509757675 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/d13a278e-db5c-44eb-9e0e-9c370731a53c/cache.tzst -P -C /home/runner/work/r-builds/r-builds
Cache restored successfully

Cache restored from key: centos-7-buildx-5d850d399d1799ab807eae39b180f063021242ae
```

The default build matrix seems to work fine, but I also want to test this with the full platform and R version matrix to ensure that the parallelization doesn't cause any issues for GHA.

shepherdjerred:
Sounds good! I wish GitHub Actions made caching a simple boolean flag haha

Issue #62, opened 2020-05-04
Some builds hang on Ubuntu 20.04 LTS


joncfoo:
@glin and I saw some builds hang on Ubuntu 20.04 LTS during the `byte-compiling grDevices` step of the build.

From what we can tell, R <= ~3.5 is affected (3.4.3, 3.5.3 were seen hanging, while 3.6.3 and 4.0.0 seemed fine).

This has _something_ to do with including `libopenblas` as a dependency - not entirely sure why the build hangs but removing this dependency allows the builds to complete.

During the _hang_, a single CPU spins at 100%.

----
Related: https://github.com/rstudio/r-builds/pull/60
Related: https://github.com/rstudio/r-builds/pull/61

glin:
This seems to have been a bug with the libopenblas package on Ubuntu 20.04: https://bugs.launchpad.net/ubuntu/+source/openblas/+bug/1860601
https://github.com/xianyi/OpenBLAS/issues/2642

Looks like it was fixed upstream last June, so we could try adding OpenBLAS again.

glin:
Fixed in https://github.com/rstudio/r-builds/pull/215 (@michaelmayer2)

PR #132, opened 2022-08-15
Openblas java

michaelmayer2:
This is an extension of https://github.com/rstudio/r-builds/pull/113 with additional stuff around OpenBLAS support (make it more consistent and also add OpenBLAS support for Ubuntu 20.04 (Ubuntu 20.04 currently only works with default BLAS/LAPACK and hence some benchmarks are 5 to 6 x slower compared to OpenBLAS. 

As for general benchmarking I have put together a github repo in https://github.com/michaelmayer2/r-builds-bench that will run https://github.com/michaelmayer2/r-benchmarks/blob/main/R-benchmark-25/R-benchmark-25.R. That's also how I found out about the performance issue in Ubuntu 20.04.

### DIFF
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -3,6 +3,7 @@ FROM centos:centos7
 ENV OS_IDENTIFIER centos-7
 
 RUN yum -y update \
+    && yum -y install epel-release \
     && yum -y install \
     autoconf \
     automake \
@@ -26,6 +27,7 @@ RUN yum -y update \
     libtool \
     make \
     ncurses-devel \
+    openblas-devel \
     pango-devel \
     pcre-devel \
     pcre2-devel \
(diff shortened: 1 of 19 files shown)


michaelmayer2:
Note: For OpenBLAS I am also doing away with the whole post install stuff in some distro's that would create symbolic links from openblas into the R installation. While it technically works I feel it is much more cleaner to do it consistently right from the start. 

(If at all we should look at flexiblas at some point but this is a story for anothe time)

glin:
Adding OpenBLAS to Ubuntu 20 sounds great. We originally had to remove it because the OpenBLAS package on Ubuntu 20 had a bug that caused R to hang, but it seems that bug is now fixed. This should close https://github.com/rstudio/r-builds/issues/62.

I think the symbolic links should be left in, however. On CentOS/RHEL and SUSE, we link R against the generic shared BLAS to make the R binaries more portable and allow users to switch out BLAS implementations without having to recompile R. We also do it to make RSPM's binary packages more portable and possible to use with the default RHEL/SUSE distribution R. We don't do this for Ubuntu/Debian because BLAS is swappable at runtime using the alternatives system there, so we link against external BLAS like Ubuntu/Debian does. There's some documentation that explains this better (which should probably be added to this repo), so I'll try to dig that up.

Issue #119, opened 2022-04-05
Existing .deb packages generate many warnings and errors from the linter


atheriel:
For an unrelated experiment I ran Debian's official package linter -- `lintian` -- on one of the generated `.deb` files. This gives:

```shell
$ lintian output/ubuntu-2204/r-4.1.3_1_amd64.deb 
E: r-4.1.3: dir-or-file-in-opt opt/R/
E: r-4.1.3: dir-or-file-in-opt opt/R/4.1.3/
E: r-4.1.3: dir-or-file-in-opt opt/R/4.1.3/bin/
E: r-4.1.3: dir-or-file-in-opt ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: r-4.1.3: extended-description-is-empty
E: r-4.1.3: malformed-contact Maintainer //github.com/rstudio/r-builds
E: r-4.1.3: malformed-contact Maintainer RStudio
E: r-4.1.3: no-copyright-file
E: r-4.1.3: too-many-contacts Maintainer RStudio, PBC https://github.com/rstudio/r-builds
E: r-4.1.3: unstripped-binary-or-object opt/R/4.1.3/bin/Rscript
E: r-4.1.3: unstripped-binary-or-object opt/R/4.1.3/lib/R/bin/Rscript
E: r-4.1.3: unstripped-binary-or-object opt/R/4.1.3/lib/R/bin/exec/R
E: r-4.1.3: unstripped-binary-or-object ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: r-4.1.3: executable-not-elf-or-script opt/R/4.1.3/lib/R/bin/BATCH
W: r-4.1.3: executable-not-elf-or-script opt/R/4.1.3/lib/R/bin/COMPILE
W: r-4.1.3: executable-not-elf-or-script opt/R/4.1.3/lib/R/bin/INSTALL
W: r-4.1.3: executable-not-elf-or-script ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: r-4.1.3: national-encoding opt/R/4.1.3/lib/R/doc/FAQ
W: r-4.1.3: package-contains-timestamped-gzip opt/R/4.1.3/lib/R/library/grDevices/afm/ArialMT-Bold.afm.gz 2022-04-05T21:05:12
[...]

PR #133, opened 2022-08-17
Add R builds for RHEL 9

glin:
Add R builds for RHEL 9, based on Rocky Linux 9. This was mostly copied over from the centos-8 (Rocky 8) build, with a few differences to note:
- The platform name is now "rhel-9" since we officially only support RHEL, not the RHEL derivatives. Rocky Linux and AlmaLinux are still mentioned in the README because the installation steps differ vs. RHEL.
- EPEL is now required to install R again, for OpenBLAS and texinfo. Rocky/Alma have replaced PowerTools with the CodeReady Builder repo as a requirement to use EPEL.
- R is now configured with Java 11 instead of Java 8 by default, as Java 8 is EOL. `JAVA_HOME` is explicitly set to a major-versioned path to make Java configuration easier on users, based on @michaelmayer2's work in https://github.com/rstudio/r-builds/pull/113.
- Some RPM fields are corrected after checking with rpmlint, like #119 suggests:
  - added https to URL, https://www.r-project.org
  - wrapped maintainer field in brackets, `RStudio, PBC <https://github.com/rstudio/r-builds>`
  - corrected the invalid `GPL-2` license field to the RPM standard `GPLv2+`
- Added more doc comments to make future maintenance easier, like how we link against BLAS.

I've run a test workflow with the full matrix of R versions at https://github.com/rstudio/r-builds/actions/runs/2878921915

### DIFF
--- a/builder/package.rhel-9
+++ b/builder/package.rhel-9
@@ -1,0 +1,60 @@
+if [[ ! -d /tmp/output/rhel-9 ]]; then
+  mkdir -p /tmp/output/rhel-9
+fi
+
+# R 3.x requires PCRE1
+pcre_lib='pcre2-devel'
+if [[ "${R_VERSION}" =~ ^3 ]]; then
+  pcre_lib='pcre-devel'
+fi
+
+# Create post-install script required for OpenBLAS
+cat <<EOF >> /post-install.sh
+mv /opt/R/${R_VERSION}/lib/R/lib/libRblas.so /opt/R/${R_VERSION}/lib/R/lib/libRblas.so.keep
+ln -s /usr/lib64/libopenblasp.so /opt/R/${R_VERSION}/lib/R/lib/libRblas.so
+EOF
+
+# Create after-remove script to remove internal BLAS
+cat <<EOF >> /after-remove.sh
+if [ -d /opt/R/${R_VERSION} ]; then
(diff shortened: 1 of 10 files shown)


builder/package.rhel-9
```diff
@@ -0,0 +1,76 @@
...
+# with the default R distributed by RHEL/SUSE, or other R installations using
+# shared BLAS and configured with a different BLAS (such as Microsoft R Open with MKL).
+# This is especially important for RSPM's binary packages.
+#
+# On Ubuntu/Debian, we link R against the external BLAS instead (--with-blas/--with-lapack),
+# as those distributions use the alternatives system to swap BLAS libraries at runtime.
+# The default R distributions on Ubuntu/Debian use the external BLAS, so we do as well
+# for portability.
+#
+# https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Shared-BLAS
```
glin, line 27: @michaelmayer2 I've added some docs to explain how we configure BLAS here (https://github.com/rstudio/r-builds/pull/132)

I'll add a similar comment to the openSUSE 15.4 build as well. I didn't copy these over to the previous CentOS versions since we usually just add new platforms based off the most recent version of that distro.

jonyoder (approved):

PR #138, opened 2022-08-26
Use latest version of the launch template

glin:
Changes to the launch template (e.g., ECS volume type/size at https://github.com/rstudio/r-builds/pull/137) were not being applied because the launch template was set to the default version 1 rather than the latest changed version.


jforest (approved):
🤞  I wonder about not putting `$Latest` in quotes, will the `$` be interpreted by cloudformation?  Would it be safer to just `'$Latest'` instead?

glin:
I'm not sure, but a quoted string sounds great regardless. I made the change.

PR #136, opened 2022-08-23
Fix openSUSE 15.4 image build; begin replacing fpm with nFPM

glin:
The SUSE 15.4 image build failed as soon as I merged https://github.com/rstudio/r-builds/pull/134 due to a Ruby gem dependency issue with fpm, the DEB/RPM packaging tool:
https://github.com/rstudio/r-builds/runs/7924309239?check_suite_focus=true
```sh
Step 6/12 : RUN gem install fpm &&     ln -s /usr/bin/fpm.ruby2.5 /usr/local/bin/fpm
 ---> Running in 6ac62b82d651
ERROR:  Error installing fpm:
	The last version of public_suffix (< 6.0, >= 2.0.2) to support your Ruby & RubyGems was 4.0.7. Try installing it with `gem install public_suffix -v 4.0.7` and then running the current command again
	public_suffix requires Ruby version >= 2.6. The current ruby version is 2.5.0.
```
The public_suffix gem depends on Ruby >= 2.6 which openSUSE doesn't provide, so we would've had to pin fpm to an older version or wait for a fix in fpm.

Instead, I've replaced fpm with a different tool called [nFPM](https://nfpm.goreleaser.com/), which @atheriel recommended a while back:
> nFPM is a simple, 0-dependencies, deb, rpm and apk packager.
>
> While [fpm](https://github.com/jordansissel/fpm) is great, for me, it is a bummer that it depends on ruby, tar and other software.
> I wanted something that could be used as a binary and/or as a library and that was really simple.
> So I decided to create nFPM: a simpler, 0-dependency, as-little-assumptions-as-possible alternative to fpm.

nFPM is a Go-based alternative to fpm that should be more reliable, has better support for code signing (https://github.com/rstudio/r-builds/issues/7), and still has all the features we need from fpm. fpm has had so many Ruby gem dependency issues over the years that have made it a pain to maintain:
https://github.com/rstudio/r-builds/issues/86
https://github.com/rstudio/r-builds/pull/110#discussion_r799052928
https://github.com/rstudio/r-builds/pull/83#issuecomment-767791975
https://github.com/rstudio/r-builds/pull/64#issuecomment-637677923

The migration was pretty straightforward, and there were just a few differences in the configs to account for:
[...]

PR #141, opened 2022-08-31
Migrate remaining platforms to nFPM

glin:
A follow-up of https://github.com/rstudio/r-builds/pull/136: replace fpm with nFPM for all other platforms. These are the same changes from https://github.com/rstudio/r-builds/pull/136 applied to the remaining platforms, so see that PR for more details. So far, the packages built using nFPM for RHEL 9 and SUSE 15.4 seem to be working just fine.


jonyoder (approved):
:+1:

Issue #149, opened 2022-12-15
Support for ARM64 builds


michaelmayer2:
It would be great if we could offer ARM64 builds for our r-builds. 

The main benefit would be the ability to leverage [AWS graviton ](https://www.anandtech.com/show/15578/cloud-clash-amazon-graviton2-arm-against-intel-and-amd/) instances with an end-to-end stack of Posit Products given the performance/price. 

I have already managed to tweak the basic build infrastructure to create ARM64 R packages but still missing is the consideration on how to store the built package in a multi-arch mode and make them accessible.

gaborcsardi:
The other use case of this running Docker (or similar) on arm64 macOS.

cap10morgan:
I was looking into this and wanted to note that GitHub Actions now has ARM64 runners available for Teams and Enterprise plans: https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/

If the team's consensus is that it would be good to move from Jenkins to GHA for the actual build & deploy of these packages, that would be a good way to do it assuming you have the appropriate account type on this GH org. If not, they're still working on making ARM64 runners available to open source repos and hoping to have that ready by the end of the year (2024).

I would be happy to work on that if it seems like a good path forward. I'm currently applying to an engineering position with Posit and want to familiarize myself with and chip in on parts of the code where I can. :)

glin:
This is now done with https://github.com/rstudio/r-builds/pull/256

The Install R documentation has been updated with steps on how to install the ARM64 builds: https://docs.posit.co/resources/install-r.html

Issue #7, opened 2019-05-14
GPG signed binaries


JsizzleR:
It would be ideal to have the binaries GPG signed and subsequently verified in the suggested shell script: https://github.com/rstudio/r-builds#example-usage

https://www.rstudio.com/code-signing/

datakid:
Requested here: https://rstudioide.zendesk.com/agent/tickets/71907

datakid:
This came up in internal ticket #71907

JsizzleR:
Requested by pro customer through support ticket, 73338

kfeinauer:
Will chat with @bdeitte about transferring this issue over to https://github.com/rstudio/platform-team/issues

atheriel:
Som digging into this yielded the following:

* [Jon Yoder reported](https://github.com/rstudio/r-builds/pull/23#discussion_r313063121) that our Jenkins environment should already contain our GPG key, but we would still need modify this repo to inject them into builds. We would need to confirm this can be done before starting on other implementation work.

* The tool we use for building RPM and DEB packages is [fpm](https://github.com/jordansissel/fpm), which supports RPM signatures by passing [`--rpm-sign`](https://fpm.readthedocs.io/en/latest/packages/rpm.html?highlight=rpm-sign) but [does not have a Debian equivalent](https://github.com/jordansissel/fpm/issues/809).

* Actually using RPM signatures requires setting up the build environment correctly. Red Hat has [support documentation on this](https://access.redhat.com/articles/3359321) but it may differ slightly for other platforms.

* I think we can still sign Debian packages using [`dpkg-sig`](https://manpages.debian.org/jessie/dpkg-sig/dpkg-sig.1.en.html) directly.

* We should also add a build step that verifies the signatures have been added correctly.

* We should start adding signature verification to internal consumers of these packages, such as [the public Docker images we build from them](https://github.com/rstudio/r-docker).

datakid:
This came up on internal ticket #74624

atheriel:
Small update on this:

* Our SRE team is not comfortable exposing our corporate GPG key in a new location (namely, AWS SecretManager), which is the major roadblock to implementing this. However, if we are able to migrate builds to GitHub Actions it will be possible to make use of the GPG keys already stored there (at the organization level).

* Signing packages with fpm is fraught, but recent efforts to migrate to nFPM (#136) will make it much easier to implement this if we ever solve the GPG key access issue.

datakid:
[...]

PR #152, opened 2023-01-10
Support arm64 builds

gaborcsardi:
This is currently experimental. Right now it does work fine on my arm64 Macbook, but probably more changes are needed for building arm64 packages on our CI and for deployment, and docs as well.

### DIFF
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -49,9 +49,11 @@ RUN yum -y update \
 RUN pip3 install awscli --upgrade --user && \
     ln -s /root/.local/bin/aws /usr/bin/aws
 
-RUN curl -LO https://github.com/goreleaser/nfpm/releases/download/v2.18.1/nfpm_amd64.rpm && \
-    yum install -y nfpm_amd64.rpm && \
-    rm nfpm_amd64.rpm
+RUN if [ "$(arch)" == "aarch64" ]; then echo arm64; else echo amd64; fi > /tmp/arch
+
+RUN curl -LO "https://github.com/goreleaser/nfpm/releases/download/v2.18.1/nfpm_$(cat /tmp/arch).rpm" && \
+    yum install -y "nfpm_$(cat /tmp/arch).rpm" && \
+    rm "nfpm_$(cat /tmp/arch).rpm"
 
 RUN chmod 0777 /opt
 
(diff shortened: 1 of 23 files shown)


gaborcsardi:
Related: #149.

gaborcsardi:
Actually, it doesn't work yet, the final deb/rpm package is not marked as arm64 currently.

gaborcsardi:
OK, it works now I checked some RPM and DEB packages manually.

gaborcsardi:
I build arm64 packages for a bunch of R versions. They are here: https://github.com/r-hub/R/releases

IDK if we want to make arm64 packages "official" and/or if we want to move them to our storage space.

Nevertheless, these packages are pretty useful for me and potentially others with an arm64 laptop.

glin (approved):
Looks good to me. I wonder why the GitHub Actions didn't run though, hmm. Probably because this is from a fork and the config doesn't have `on.pull_request`, https://github.com/rstudio/r-builds/blob/1a4f8a8fb3f0a3787e4ab6a54d4268791edfde05/.github/workflows/test.yml#L3-L4. I'll run it here just to double check the amd64 builds: https://github.com/rstudio/r-builds/compare/feature/arm64-pr?expand=1

For making the arm64 packages official, we don't have it in our platform support yet, but I think it's definitely coming as customer requests have been picking up, and IDE/Workbench is starting to provide arm64 builds now. 

As for CI and deployment, I think there are more open questions around that. I'd guess that getting arm64 builds into the existing AWS infrastructure would work, but take some effort to figure out. We'll need to build the Docker images for arm64 in Jenkins, and then get Batch jobs running on the arm64 graviton instances probably. Or perhaps we want to start migrating to GitHub Actions (e.g. for https://github.com/rstudio/r-builds/issues/7), which I bet would be much easier as the docker actions already support cross-platform building.

For storage, the existing S3 buckets seem fine, and I believe we're also going to eventually (soon?) push these binaries to DEB/RPM repos on Cloudsmith.

gaborcsardi:
> which I bet would be much easier as the docker actions already support cross-platform building.

It is indeed not too bad to set this up, e.g. https://github.com/r-hub/r-minimal has arm64 builds: 
https://github.com/r-hub/r-minimal/actions/runs/3899975536/workflow#L58-L62
https://github.com/r-hub/r-minimal/actions/runs/3899975536/workflow#L81-L88
and this already pushes to Docker hub and ghcr.io:
https://github.com/r-hub/r-minimal/actions/runs/3899975536/workflow#L45-L56

The only issue is that they are very slow, it takes almost an hour to build this very minimal version of R. Nevertheless it probably makes sense to set up a GHA workflow to do the daily `next` and `devel` arm64 builds on GHA and then publish them at https://github.com/r-hub/R/releases for now.

gaborcsardi:
@glin Thanks for the review!

Issue #153, opened 2023-01-11
R-devel does not compile on Ubuntu 18.04, possibly others


gaborcsardi:
The last R-devel build for Ubuntu 18.04 is from 2023-01-04, possibly because of this: https://github.com/wch/r-source/commit/71136eccddac0e836c87ea7b77b5fc43d48cbd78

I _think_ this affects (only tested Ubuntu 18.04 explicitly):
- Ubuntu 18.04 (EOL at April 30 2023),
- CentOS 7 (maintenance updates until 2024-06-30)
- OpenSUSE 15.3 (EOL already at the end of 2022)
- OpenSUSE 15.4 (EOL at end of November 2023)

Compilation fails like this on Ubuntu 18.04:
```
* installing *source* package 'MASS' ...
** package 'MASS' successfully unpacked and MD5 sums checked
** using non-staged installation
** libs
using C compiler: 'gcc (Ubuntu/Linaro 7.5.0-3ubuntu1~18.04) 7.5.0'
make[3]: Entering directory '/tmp/RtmpjT1xnC/R.INSTALL90284ed01b1b/MASS/src'
gcc -std=gnu17 -I"/root/R-devel/include" -DNDEBUG   -I/usr/local/include    -fPIC  -g -O2  -c MASS.c -o MASS.o
gcc: error: unrecognized command line option '-std=gnu17'; did you mean '-std=gnu11'?
/root/R-devel/etc/Makeconf:184: recipe for target 'MASS.o' failed
make[3]: *** [MASS.o] Error 1
make[3]: Leaving directory '/tmp/RtmpjT1xnC/R.INSTALL90284ed01b1b/MASS/src'
ERROR: compilation failed for package 'MASS'
```

Should we do anything about this? Should we email R-core? Or patch R ourselves?

The new version of R will probably come out around the end of April, so for that only CentOS 7 (well, maintenance), and OpenSUSE 15.4 will be affected.

glin:
From the commit message, "force C17 until mgcv gets updated", do you think it'll be a temporary failure? If not, then yeah that seems quite bad to have so many still supported distros unable to compile.

I had also been looking at another R-devel failure with openSUSE 15.3/15.4 (and probably CentOS 7) a while back, and hoping that was a bug that would be fixed at some point. But since we're getting close to April, it might be time to look more into it and ask about it as well. I wonder if you've also seen or know about a compilation error like this, seemingly related to the change with the internally-included BLAS/LAPACK:
```sh
[...]

PR #156, opened 2023-02-02
Use gcc-8 on Ubuntu 18.04

gaborcsardi:
To address #153 on Ubuntu 18.04.

If this the way we are going, then we can do a similar exception for OpenSUSE. CentOS 7 is a bit trickier because it needs an extra repository first.


gaborcsardi:
Need to fix the package file, gcc-8 must be conditional.

gaborcsardi:
@glin Seems to be OK now.

glin (approved):

gaborcsardi:
Thanks! I also did the OpenSUSE fix, I'll submit that as well.

PR #157, opened 2023-02-07
Fix R-devel compilation on OpenSUSE

gaborcsardi:
By using newer compilers.

### DIFF
--- a/builder/Dockerfile.opensuse-153
+++ b/builder/Dockerfile.opensuse-153
@@ -92,4 +92,5 @@ ENV CONFIGURE_OPTIONS="\
 
 COPY package.opensuse-153 /package.sh
 COPY build.sh .
+COPY patches /patches
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.opensuse-153
+++ b/builder/Dockerfile.opensuse-153
@@ -15,6 +15,9 @@ RUN zypper --non-interactive --gpg-auto-import-keys -n install \
     gcc \
     gcc-c++ \
     gcc-fortran \
+    gcc8 \
+    gcc8-c++ \
+    gcc8-fortran \
     glib2-devel \
     glibc-locale \
     help2man \
--- a/builder/Dockerfile.opensuse-154
+++ b/builder/Dockerfile.opensuse-154
@@ -102,4 +102,5 @@ ENV JAVA_HOME=/usr/lib64/jvm/jre-11-openjdk
 
 COPY package.opensuse-154 /package.sh
 COPY build.sh .
+COPY patches /patches
 ENTRYPOINT ./build.sh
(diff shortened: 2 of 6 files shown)


gaborcsardi:
Needs a fix, obviously.

glin (approved):
Looks good, but are you getting this BLAS/LAPACK compilation error that I ran into a while back? https://github.com/rstudio/r-builds/issues/153#issuecomment-1379540584
```sh
gcc-8 -shared -fopenmp -L/usr/local/lib -o libRlapack.so dlamch.o dlapack.o cmplx.o dlartg.o dlassq.o la_constants.o la_xisnan.o zlartg.o zlassq.o   -L"../../../lib" -lRblas -lgfortran -lm -lquadmath 
/usr/lib64/gcc/x86_64-suse-linux/8/../../../../x86_64-suse-linux/bin/ld: zlassq.o: warning: relocation against `__la_xisnan_MOD_disnan' in read-only section `.text'
/usr/lib64/gcc/x86_64-suse-linux/8/../../../../x86_64-suse-linux/bin/ld: dlassq.o: relocation R_X86_64_PC32 against symbol `__la_xisnan_MOD_disnan' can not be used when making a shared object; recompile with -fPIC
/usr/lib64/gcc/x86_64-suse-linux/8/../../../../x86_64-suse-linux/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make[4]: *** [Makefile:104: libRlapack.so] Error 1
```
I still hit that error when testing these changes on openSUSE 15.3 and 15.4. I was hoping this would be fixed with the newer gcc, but it seems like a different issue.

gaborcsardi:
Interesting, I was actually using this patch for the arm64 builds, and there are no errors there... seems like an x86_64 issue maybe?

gaborcsardi:
Also, I don't really follow what is linking to what when the error happens, which library needs to be recompiled with `-fPIC`?

gaborcsardi:
It is a bug in R, I think. I reported it, but it is also not hard to fix it, I believe.

gaborcsardi:
Let's wait a couple of days with this, if they fix the bug in R-core, then we can just merge it as it. Otherwise we can patch the R sources to fix it.

glin:
Sounds good, thanks for looking into that!

gaborcsardi:
OK, this should fix the devel build as well. The patch is applied on all platforms, but it is the right thing to do, anyway.

Btw. do we want to run the 'next' and 'devel' builds on the CI for PRs?

glin (approved):

glin:
About running next and devel builds on CI, sure, that sounds like a good idea. I didn't add devel originally because it could be failing for unrelated reasons, but I guess that shouldn't hold up merging PRs anyway. It'd be really useful for testing changes like this.

gaborcsardi:
Thanks for reviewing this!

PR #158, opened 2023-02-10
Fix R-devel build on CentOS 7

gaborcsardi:
- need to compile a newer (static) libcurl.
- need to set CC17, because recommended packages are now installed with a CC17
  compiler, although this is only temporary. The CC17 env var is not used by
  previous versions of R, only current R-devel.

One glitch is that we still depend on the libcurl-devel system package, and this is not necessary any more. I checked that R does not link to the system libcurl any more, and also reports the correct version for `libcurlVersion()`.

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -64,6 +64,17 @@ fetch_r_source() {
   rm /tmp/R-${1}.tar.gz
 }
 
+install_libcurl() {
+    cd /tmp
+    curl -LO https://curl.se/download/curl-7.32.0.tar.gz
+    tar xzf curl-*.tar.gz
+    yum install -y zlib-devel openssl-devel make groff
+    cd curl-7.32.0
+    ./configure --disable-shared --with-pic
+    make
+    make install
+}
+
 # compile_r() - $1 as r version
 compile_r() {
   cd /tmp/R-${1}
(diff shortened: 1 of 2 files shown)


builder/build.sh
```diff
@@ -64,6 +64,17 @@ fetch_r_source() {
   rm /tmp/R-${1}.tar.gz
 }
 
+install_libcurl() {
```
glin, line 67: A comment here (or in `set_up_environment`) that explains why we do this would be nice for future reference, even if we'll end up removing it in a year anyway.

glin: Or alternatively, download and compile this in the CentOS 7 image (but don't install it unless R >= 4.3)? Precompiling might shave off some build time as well.

glin (approved):
Looks great, just added a small suggestion. I also tested this with the patch from https://github.com/rstudio/r-builds/pull/157 and confirmed that it fixes the LAPACK issue on x86_64 as well.

builder/build.sh
gaborcsardi: We could do that, but downloading and compiling libcurl takes about less than two  minutes, so I am not sure if it is worth it.

gaborcsardi: In the end I did change it to build from the centos-7 Dockerfile, and only install in the build script on CentOS7 and only for R-devel.

The other thing that I'll still try is installing a newer libcurl version. AFAIK RedHat/CentOS backports bug fixes, so a stock 7.32.0 might be worse than a CentOS 7 7.29.x security-wise.

gaborcsardi: OK, the latest libcurl version seems to work.

gaborcsardi:
So this is not needed any more, I guess?

PR #159, opened 2023-02-27
Feature sns

stevenolen:
Our team has a [need/interest](https://github.com/rstudio/shinyapps-users-guide/issues/80) in ensuring we are kept up-to-date on new versions of r-builds being published.

This adds an sns topic and step to the step functions to publish a list of version/platform pairs which were successfully built during the active run (and does not publish if there were no successful jobs). I refactored the batch job object interaction a little, I was having trouble following our past logic 😄 .

I also bumped the serverless node deps... they were outlandishly old and we have a few other projects that have been using these versions for awhile!

### DIFF
--- a/handler.py
+++ b/handler.py
@@ -8,6 +8,27 @@ import botocore
 CRAN_SRC_R3_URL = 'https://cran.rstudio.com/src/base/R-3/'
 CRAN_SRC_R4_URL = 'https://cran.rstudio.com/src/base/R-4/'
 batch_client = boto3.client('batch', region_name='us-east-1')
+sns_client = boto3.client('sns', region_name='us-east-1')
+
+
+class JobDetails:
+    def __init__(self, version, platform):
+        self.version = version
+        self.platform = platform
+
+    @classmethod
+    def from_job_name(cls, job_name):
+        _R, version, platform = job_name.split('-', 2)
+        return cls(version.replace('_', '.'), platform)
+
+    def job_name(self):
+        return f"R-{self.version.replace('.', '_')}-{self.platform}"
+
+    def job_definition_arn(self):
+        return os.environ[f"JOB_DEFINITION_ARN_{self.platform.replace('-','_')}"]
+
+    def to_json(self):
+        return {'version': self.version, 'platform': self.platform}
 
 
 def _to_list(input):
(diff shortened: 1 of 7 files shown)


stevenolen:
only failed job was centos7/r4.0.5, for an unrelated reason (i believe):

```
---> Package R-4.0.5.x86_64 0:1-1 will be installed
--> Processing Dependency: openblas-devel for package: R-4.0.5-1-1.x86_64
---> Package libpciaccess.x86_64 0:0.14-1.el7 will be installed
--> Processing Dependency: hwdata for package: libpciaccess-0.14-1.el7.x86_64
--> Running transaction check
---> Package R-4.0.5.x86_64 0:1-1 will be installed
--> Processing Dependency: openblas-devel for package: R-4.0.5-1-1.x86_64
---> Package hwdata.x86_64 0:0.252-9.7.el7 will be installed
--> Finished Dependency Resolution
Error: Package: R-4.0.5-1-1.x86_64 (/R-4.0.5-1-1.x86_64)
 You could try using --skip-broken to work around the problem
           Requires: openblas-devel
 You could try running: rpm -Va --nofiles --nodigest
```

handler.py
```diff
@@ -145,3 +165,29 @@ def poll_running_jobs(event, context):
...
+    """Publish details about successfully finished jobs."""
+
+    # first, if there were no succeeded jobs, return instead of publishing details about builds
+    if len(event['succeededJobIds']) < 1:
+        return event
+
+    # fetch all jobs, removing those which are not in our succeeded id list
+    r = batch_client.list_jobs(jobQueue=os.environ['JOB_QUEUE_ARN'], jobStatus='SUCCEEDED')
+    print(f'r: {r}')
+    succeeded_jobs = [i for i in r['jobSummaryList'] if i['jobId'] in event['succeededJobIds']]
```
jforest, line 180: Maybe this is a stupid question, but is there a case when a job status would be `SUCCEEDED` but not in `event['succeededJobIds']`?  Isn't the `r` data structure already purely `SUCCEEDED` jobs?

package.json
```diff
@@ -4,9 +4,9 @@
   "version": "0.1.0",
   "dependencies": {},
   "devDependencies": {
-    "serverless": "^1.42.3",
+    "serverless": "^3.21.0",
```
jforest, line 7: slight upgrade

serverless-stepf.yml
```diff
@@ -1,5 +1,4 @@
 rBuilds:
-  Comment: "R Builds"
```
jforest, line 2: @stevenolen hates comments?

jforest (approved):

serverless-stepf.yml
stevenolen: oh, they took this out of the spec, i had forgotten i made this change!

PR #161, opened 2023-04-19
Remove R-devel patch for openSUSE; add R-devel to CI tests

glin:
- Remove patches now that the openSUSE issue is fixed upstream (https://github.com/wch/r-source/commit/81607372bca0b0eca7826d3f20ad9047f7016a1d) (https://github.com/rstudio/r-builds/pull/157). Keep the patching mechanism however, as it seems like it'll be useful in the future.
- Add R-devel to CI tests (https://github.com/rstudio/r-builds/pull/157#issuecomment-1425890015). I didn't add `next` though, as it's the same as `devel` most of the year.


gaborcsardi:
Looks good! The openblas-devel CI failure is hopefully temporary, but it is not related to this PR anyway.

I agree that testing `next` is not important, it is actually R-patched most of the year, which is almost the same as R-release.

gaborcsardi (approved):

Issue #165, opened 2023-04-20
R-builds deployment failing


glin:
Looks like the deploy job started failing yesterday, but no obvious reason in the error messages. This should be fixed asap since we have some changes to deploy for R 4.3.0 tomorrow.

```sh
[2023-04-19T18:37:26.606Z] Error: `docker run --rm -v .cache/serverless-python-requirements/2b10c0467f75bfbcf12e9b77be65ec7563b90c5995f0eaa96faefbd42c337bdd_x86_64_slspyc:/var/task:z -v .cache/serverless-python-requirements/downloadCacheslspyc:/var/useDownloadCache:z lambci/lambda:build-python3.8 /bin/sh -c chown -R 0\:0 /var/useDownloadCache && python3.8 -m pip install -t /var/task/ -r /var/task/requirements.txt --cache-dir /var/useDownloadCache && chown -R 386\:386 /var/task && chown -R 386\:386 /var/useDownloadCache` Exited with code 125
[2023-04-19T18:37:26.606Z]     at ChildProcess.<anonymous> (/var/lib/jenkins/workspace/r-builds/deploy-r-builds/node_modules/child-process-ext/spawn.js:38:8)
[2023-04-19T18:37:26.606Z]     at ChildProcess.emit (events.js:400:28)
[2023-04-19T18:37:26.606Z]     at ChildProcess.emit (domain.js:475:12)
[2023-04-19T18:37:26.606Z]     at maybeClose (internal/child_process.js:1088:16)
[2023-04-19T18:37:26.606Z]     at Socket.<anonymous> (internal/child_process.js:446:11)
[2023-04-19T18:37:26.606Z]     at Socket.emit (events.js:400:28)
[2023-04-19T18:37:26.606Z]     at Socket.emit (domain.js:475:12)
[2023-04-19T18:37:26.606Z]     at Pipe.<anonymous> (net.js:686:12)
[2023-04-19T18:37:26.606Z] 
[2023-04-19T18:37:26.606Z] 2 deprecations found: run 'serverless doctor' for more details
[2023-04-19T18:37:26.606Z] make: *** [Makefile:40: serverless-deploy.staging] Error 1
```

glin:
[...]

PR #166, opened 2023-06-17
Add support for Debian 12

gaborcsardi:


glin (approved):
Looks good, thanks for adding this. Unfortunately I think we're still blocked on deploying infra changes like this though, because of https://github.com/rstudio/r-builds/issues/165. But we could try to deploy again and see what happens with Jenkins.

gaborcsardi:
Thanks!

PR #171, opened 2023-07-05
Fix R-builds deploy job

### LINKED ISSUE #165: R-builds deployment failing

glin:
Fixes #165. Tested this in staging at https://build.posit.it/blue/organizations/jenkins/r-builds%2Fdeploy-r-builds/detail/deploy-r-builds/195/pipeline, and then confirmed that a staging rebuild works. The CI failure is for an unrelated reason.

If this works well, someone will need to update serverless-custom.yml since I don't think I have permissions for that.

We were using the serverless-python-requirements plugin with `dockerizePip: true`, which installs the Python requirements in a separate docker container. The plugin has to mount some directories (requirements.txt, cache files) into the separate container, and this failed in two ways.

The first issue is that the mounted directory was ending up as a relative path starting with a period, which docker doesn't support. With `serverless --verbose`, you get the actual error message:
```sh
[2023-07-04T00:11:49.031Z] + docker run --rm -v .cache/serverless-python-requirements/2b10c0467f75bfbcf12e9b77be65ec7563b90c5995f0eaa96faefbd42c337bdd_x86_64_slspyc:/var/task:z -v .cache/serverless-python-requirements/downloadCacheslspyc:/var/useDownloadCache:z lambci/lambda:build-python3.8 /bin/sh -c chown -R 0:0 /var/useDownloadCache
[2023-07-04T00:11:49.031Z] docker: Error response from daemon: create .cache/serverless-python-requirements/2b10c0467f75bfbcf12e9b77be65ec7563b90c5995f0eaa96faefbd42c337bdd_x86_64_slspyc: ".cache/serverless-python-requirements/2b10c0467f75bfbcf12e9b77be65ec7563b90c5995f0eaa96faefbd42c337bdd_x86_64_slspyc" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
[2023-07-04T00:11:49.031Z] See 'docker run --help'.
script returned exit code 125
```
This cache dir comes from the plugin's use of the `appdirectory` npm package, which generates the cache paths based off `$HOME`: https://github.com/MrJohz/appdirectory/blob/27f19a6eceb46110cd5d6882a18cae3a4da98331/lib/appdirectory.js#L91-L94

When running the plugin locally in Docker, it works fine because `$HOME` resolves to `/root` or `/home/<user>`. However, in Jenkins, the docker image is run with `HOME` set to `.`:
```sh
# Docker agent in Jenkins
[2023-07-05T18:36:45.755Z] echo "HOME: $HOME"
[2023-07-05T18:36:45.755Z] HOME: .

# Docker running locally
$ echo HOME: $HOME
HOME: /root
```
~~This may have been a Jenkins behavior change in some upgrade, which could explain why the job suddenly started failing.~~ We were settting `HOME=.`, see https://github.com/rstudio/r-builds/pull/171#issuecomment-1622645902

The second issue is that we were mounting the docker socket in the agent container. When you do this, any `docker run -v` mounts from within the container will use paths from the host rather than the container. So even with the cache directory fixed, the plugin was trying to mount directories in the python container that didn't exist, causing a new issue. I'm not sure how the Jenkins job was working before. 
```sh
docker run --rm -v /root/.cache/2b10c0467f75bfbcf12e9b77be65ec7563b90c5995f0eaa96faefbd42c337bdd_x86_64_slspyc:/var/task:z -v /root/.cache/downloadCacheslspyc:/var/useDownloadCache:z lambci/lambda:build-python3.8 /bin/sh -c chown -R 0\:0 /var/useDownloadCache && python3.8 -m pip install -t /var/task/ -r /var/task/requirements.txt --cache-dir /var/useDownloadCache && chown -R 0\:0 /var/task && chown -R 0\:0 /var/useDownloadCache
chown: missing operand
```

To fix both of these, I temporarily patched serverless-custom.yml to not dockerize pip ~~and disable caching~~:
```yml
pythonRequirements:
  dockerizePip: false
```

Switched the deploy image to a base OS image rather than a python/node or lambda image, because the language specific images felt hard to maintain. The nodesource install script didn't work on the python/node images' Debian 11, and the lambda images use AL2 which is super old and no longer supports newer versions of Node.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,18 @@ ecr-login:
 push-serverless-custom-file:
 	aws s3 cp serverless-custom.yml s3://rstudio-devops/r-builds/serverless-custom.yml
 
+# Temporarily patch serverless-custom.yml to not Dockerize pip and disable
+# serverless-python-requirements caching. The caching does not work in Jenkins
+# by default because the cache directory is resolved to a relative ".cache"
+# directory, rather than an absolute directory, which breaks the plugin.
+# The cache directory must either be an absolute path (optionally specified
+# via cacheLocation: /path/to/cache), or caching must be disabled for the plugin
+# to work. We disable caching completely to avoid all sorts of future issues.
+# The cache directory is based on $HOME, which is set to "." when running
+# Docker images in Jenkins for some reason.
 fetch-serverless-custom-file:
 	aws s3 cp s3://rstudio-devops/r-builds/serverless-custom.yml .
+	sed -i 's|dockerizePip: true|dockerizePip: false\n  useStaticCache: false\n  useDownloadCache: false|' serverless-custom.yml
 
 rebuild-all: deps fetch-serverless-custom-file
 	$(SLS_BINARY) invoke stepf -n rBuilds -d '{"force": true}'
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,11 @@ rebuild-all: deps fetch-serverless-custom-file
 serverless-deploy.%: deps fetch-serverless-custom-file
 	$(SLS_BINARY) deploy --stage $* --verbose
 
+# Package the service only, for debugging.
+# Requires deps and fetch-serverless-custom-file to be run first.
+serverless-package:
+	$(SLS_BINARY) package --verbose
+
 define GEN_TARGETS
 docker-build-$(platform):
 	@cd builder && docker-compose build $(platform)
--- a/Makefile
+++ b/Makefile
@@ -30,18 +30,11 @@ ecr-login:
 push-serverless-custom-file:
 	aws s3 cp serverless-custom.yml s3://rstudio-devops/r-builds/serverless-custom.yml
 
-# Temporarily patch serverless-custom.yml to not Dockerize pip and disable
-# serverless-python-requirements caching. The caching does not work in Jenkins
-# by default because the cache directory is resolved to a relative ".cache"
-# directory, rather than an absolute directory, which breaks the plugin.
-# The cache directory must either be an absolute path (optionally specified
-# via cacheLocation: /path/to/cache), or caching must be disabled for the plugin
-# to work. We disable caching completely to avoid all sorts of future issues.
-# The cache directory is based on $HOME, which is set to "." when running
-# Docker images in Jenkins for some reason.
+# Temporarily patch serverless-custom.yml to not Dockerize pip for the
+# serverless-python-requirements plugin.
 fetch-serverless-custom-file:
 	aws s3 cp s3://rstudio-devops/r-builds/serverless-custom.yml .
-	sed -i 's|dockerizePip: true|dockerizePip: false\n  useStaticCache: false\n  useDownloadCache: false|' serverless-custom.yml
+	sed -i 's|dockerizePip: true|dockerizePip: false|' serverless-custom.yml
 
 rebuild-all: deps fetch-serverless-custom-file
 	$(SLS_BINARY) invoke stepf -n rBuilds -d '{"force": true}'
--- a/deploy.Jenkinsfile
+++ b/deploy.Jenkinsfile
@@ -3,10 +3,6 @@ pipeline {
     dockerfile {
       dir 'jenkins'
       label 'docker'
-      // Required for the serverless-python-requirements plugin to install
-      // Python requirements using the AWS build images, running Docker inside Docker.
-      additionalBuildArgs '--build-arg DOCKER_GID=$(stat -c %g /var/run/docker.sock) --build-arg JENKINS_UID=$(id -u jenkins) --build-arg JENKINS_GID=$(id -g jenkins)'
-      args '-v /var/run/docker.sock:/var/run/docker.sock --group-add docker'
     }
   }
   environment {
--- a/deploy.Jenkinsfile
+++ b/deploy.Jenkinsfile
@@ -5,13 +5,6 @@ pipeline {
       label 'docker'
     }
   }
-  environment {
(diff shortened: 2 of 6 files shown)


glin:
@jspiewak pointed out that we explicitly set `HOME` to `.` in the Jenkins job: https://github.com/rstudio/r-builds/blob/c97f0cc4dd4d275121796c9e4271ffd9bfe7372d/deploy.Jenkinsfile#L11

That explains the setting then, and how it was able to work before. The HOME dir is used for the cache, which exists on the host because it's the workspace dir that gets mounted in. Changing it to `env.WORKSPACE` as @jspiewak suggested gets the original job working again without all these other changes.

I'll still propose running the whole deployment in a single container though, to keep things simpler and easier to debug locally.

gaborcsardi (approved):
I can't really check if the changes are OK, just saying thanks for fixing the deployment. :)

Makefile
```diff
@@ -30,14 +30,29 @@ ecr-login:
...
 	aws s3 cp serverless-custom.yml s3://rstudio-devops/r-builds/serverless-custom.yml
 
+# Temporarily patch serverless-custom.yml to not Dockerize pip and disable
+# serverless-python-requirements caching. The caching does not work in Jenkins
+# by default because the cache directory is resolved to a relative ".cache"
+# directory, rather than an absolute directory, which breaks the plugin.
+# The cache directory must either be an absolute path (optionally specified
+# via cacheLocation: /path/to/cache), or caching must be disabled for the plugin
+# to work. We disable caching completely to avoid all sorts of future issues.
+# The cache directory is based on $HOME, which is set to "." when running
```
jspiewak, line 40: No longer the case

Makefile
```diff
@@ -30,14 +30,29 @@ ecr-login:
...
 rebuild-all: deps fetch-serverless-custom-file
 	$(SLS_BINARY) invoke stepf -n rBuilds -d '{"force": true}'
 
 serverless-deploy.%: deps fetch-serverless-custom-file
-	$(SLS_BINARY) deploy --stage $*
+	$(SLS_BINARY) deploy --stage $* --verbose
+
+# Package the service only, for debugging.
+# Requires deps and fetch-serverless-custom-file to be run first.
+serverless-package:
```
jspiewak, line 54: Why not declare `deps` and `fetch-serverless-custom-file` as target dependencies then?

jspiewak (commented):
If setting `HOME` to `env.WORKSPACE` actually fixes the issue, is it worth separating that fix from the rest and getting it landed?
Otherwise, @stevenolen or @jforest will need to chime in with some historical context on the docker-in-docker situation.

glin:
@jspiewak Good idea, let's go with that first to get the deploy unblocked asap. I'll follow up with a new PR soon and add some docs. The main problem was just how hard the deploy job was to understand and debug, and I would be fine with the original docker-inside-docker approach as long as it was documented well enough. I'd still prefer reducing it to just one container, but am curious what @stevenolen and @jforest think.

Makefile
glin: This was to iterate on the packaging (without deploying) step to debug failures with that. I was testing various serverless-python-requirements settings in serverless-custom.yml and didn't want to overwrite it.

glin:
I updated the docs and tried to simplify the job a little more:
- Eliminated the custom `HOME` config - restored the jenkins user so `HOME` wouldn't be `/` and break npm
- Re-enabled plugin caching so only `dockerizePip: true` would need to be removed in serverless-custom.yml

https://github.com/rstudio/r-builds/pull/175 seems to be working fine for now though, but I'll still leave this up for a while.

glin:
No longer needed

PR #174, opened 2023-07-05
Fix AWS CLI on PATH for Debian 12; install v2

### LINKED ISSUE #172: Debian 12 builds failing because AWS CLI isn't on PATH
The Debian 12 builds currently fail because AWS CLI isn't on the PATH:
```sh
./build.sh: line 22: aws: command not found
```
It's also aws cli v1, so we could update to v2 as well:
```sh
$ /root/.local/pipx/venvs/awscli/bin/aws --version
aws-cli/1.27.160 Python/3.11.2 Linux/5.14.0-1056-oem botocore/1.29.160
```
This is only when publishing, so we didn't notice it before.

glin:
Fixes #172. `pipx ensurepath` seems to just update PATH so the changes weren't saved in subsequent layers. That was also installing AWS CLI v1, so we might as well install v2 using Amazon's installer.


gaborcsardi (approved):
Thanks!

PR #176, opened 2023-07-10
Fix R <= 4.2 builds for SUSE 15.4; patch configure to allow libcurl 8.x

### LINKED ISSUE #173: openSUSE 15.4 builds failing because libcurl was updated to new major version 8
The openSUSE 15.4 R builds currently fail for R <= 4.2 because SUSE 15.4 updated libcurl from 7.x to 8.x, and R <= 4.2 needs libcurl to be 7.x:

```sh
checking libcurl version ... 8.0.1
checking for curl/curl.h... yes
checking if libcurl is version 7 and >= 7.28.0... no
configure: error: libcurl >= 7.28.0 library and headers are required with support for https

$ zypper info libcurl4
Loading repository data...
Reading installed packages...

Information for package libcurl4:
---------------------------------
Repository     : Update repository with updates from SUSE Linux Enterprise 15
Name           : libcurl4
Version        : 8.0.1-150400.5.23.1
```

R 4.2 seems to work fine with libcurl 8 installed (upgrading libcurl on an existing R image)? So we could possibly patch the configure script.
```sh
> libcurlVersion()
[1] "8.0.1"
attr(,"ssl_version")
[1] "OpenSSL/1.1.1l-fips"
attr(,"libssh_version")
[1] "libssh/0.9.6/openssl/zlib"
attr(,"protocols")
 [1] "dict"    "file"    "ftp"     "ftps"    "gopher"  "gophers" "http"   
 [8] "https"   "imap"    "imaps"   "ldap"    "ldaps"   "mqtt"    "pop3"   
[15] "pop3s"   "rtsp"    "scp"     "sftp"    "smb"     "smbs"    "smtp"   
[22] "smtps"   "telnet"  "tftp"   
```

glin:
Fixes #173. Apparently curl 8.x was a major version bump with no API/ABI changes, so that explains why SUSE upgraded it through a security update.

The `configure` script in R 4.2.3 and below bails out when the libcurl version is > 7, and R 4.3.0 changed this to allow versions > 7:

> libcurl 8.x is now accepted by configure: despite a change in major version number it changes neither API nor ABI.

https://github.com/wch/r-source/commit/da6638896413bcbb5970b2335b92582853f94e3c

This applies the same change to `configure` for all R versions <= 4.2, since newer distros will probably start coming with libcurl 8 now. The original check was added in R 3.2 at some point, and doesn't seem to have changed between then and 4.2.3: https://github.com/wch/r-source/commit/91b04408ee35a059959780c26074457749490a74 (although it seems to have been reverted in 3.2.2, then restored in 3.3.0 for some reason). So I didn't bother with putting a lower bound on the version check.

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -75,12 +75,13 @@ patch_r() {
 
 # compile_r() - $1 as r version
 compile_r() {
-  cd /tmp/R-${1}
+  r_version=${1}
+  cd "/tmp/R-${r_version}"
 
   # tools/config.guess in R versions older than 3.2.2 guess 'unknown' instead of 'pc'
   # test the version and properly set the flag.
   build_flag="--build=$(uname -m)-pc-linux-gnu"
-  if _version_is_greater_than ${R_VERSION} 3.2.2; then
+  if _version_is_greater_than "${r_version}" 3.2.2; then
     build_flag=''
   fi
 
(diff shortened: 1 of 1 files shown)


gaborcsardi (approved):
Looks great!

PR #177, opened 2023-07-10
Discontinue openSUSE 15.3 and Ubuntu 18.04 builds for EOL

### LINKED ISSUE #167: Ubuntu 18.04 does not have standard support any more, should we remove it?
According to https://wiki.ubuntu.com/Releases

glin:
Resolves #167. Both SUSE 15.3 and Ubuntu 18.04 are now EOL and can be removed.


gaborcsardi (approved):
Great, thanks!

Issue #76, opened 2020-09-09
Get R 3.3.0, 3.3.1, 3.3.2 builds working for all distros


glin:
R 3.3.0, 3.3.1, and 3.3.2 builds don't work for most of the newer distros, like Ubuntu 18 and CentOS 8, due to a bug in the configure script (see https://github.com/rstudio/r-builds/pull/48): https://github.com/rstudio/r-builds/blob/0eaf4ff908f5944c152f89128bbffb3f73ce1c95/handler.py#L128-L139

It has caused some pains with documentation and the quick install script (https://github.com/rstudio/r-builds/issues/36). @jimhester suggests we could just backport the patch for this bug, since the fix is pretty minor anyway: https://stat.ethz.ch/pipermail/r-devel/2017-February/073752.html

gaborcsardi:
Closed by #179 I believe.

Issue #36, opened 2019-09-20
Quick Install lists unavailable versions


ricardofandrade:
On Ubuntu18, running the script shows:
```
Available Versions
  3.6.1
  3.6.0
  3.5.3
  3.5.2
  3.5.1
  3.5.0
  3.4.4
  3.4.3
  3.4.2
  3.4.1
  3.4.0
  3.3.3
  3.3.2
  3.3.1
  3.3.0
  3.2.5
  3.2.4
  3.2.3
  3.2.2
  3.2.1
  3.2.0
  3.1.3
  3.1.2
  3.1.1
  3.1.0
  3.0.3
  3.0.2
  3.0.1
  3.0.0
Enter version to install: (<ENTER> for latest)
```

However it fails for 3.3.0:
```
Enter version to install: (<ENTER> for latest)
3.3.0
Downloading installer from https://cdn.rstudio.com/r/ubuntu-1804/pkgs/r-3.3.0_1_amd64.deb...
Downloading https://cdn.rstudio.com/r/ubuntu-1804/pkgs/r-3.3.0_1_amd64.deb...
--2019-09-20 16:29:40--  https://cdn.rstudio.com/r/ubuntu-1804/pkgs/r-3.3.0_1_amd64.deb
Resolving cdn.rstudio.com (cdn.rstudio.com)... 143.204.158.105, 143.204.158.126, 143.204.158.84, ...
Connecting to cdn.rstudio.com (cdn.rstudio.com)|143.204.158.105|:443... connected.
HTTP request sent, awaiting response... 403 Forbidden
2019-09-20 16:29:40 ERROR 403: Forbidden.


We were unable to download the installation bundle.
```

Similarly, in Fedora28 any version fails:
```
Enter version to install: (<ENTER> for latest)
3.5.3
Downloading installer from https://cdn.rstudio.com/r/centos-8/pkgs/R-3.5.3-1-1.x86_64.rpm...
Downloading https://cdn.rstudio.com/r/centos-8/pkgs/R-3.5.3-1-1.x86_64.rpm...
--2019-09-20 16:31:45--  https://cdn.rstudio.com/r/centos-8/pkgs/R-3.5.3-1-1.x86_64.rpm
Resolving cdn.rstudio.com (cdn.rstudio.com)... 143.204.158.92, 143.204.158.126, 143.204.158.84, ...
Connecting to cdn.rstudio.com (cdn.rstudio.com)|143.204.158.92|:443... connected.
[...]

PR #179, opened 2023-07-19
Fix R 3.3.0-3.3.2 builds

### LINKED ISSUE #178: Some builds are missing for Debian 12
Namely: 3.3.2, 3.3.1, 3.3.0. On the arm64 builds I see this:
```
debian-12_1     | checking for zlib.h... yes
debian-12_1     | checking if zlib version >= 1.2.5... no
debian-12_1     | checking whether zlib support suffices... configure: error: zlib library and headers are required
builder_debian-12_1 exited with code 1
```

gaborcsardi:
Closes #178.

### DIFF
--- a/builder/Dockerfile.centos-7
+++ b/builder/Dockerfile.centos-7
@@ -73,4 +73,5 @@ ENV R_RD4PDF="times,hyper"
 
 COPY package.centos-7 /package.sh
 COPY build.sh .
+COPY patches /patches
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.centos-8
+++ b/builder/Dockerfile.centos-8
@@ -74,4 +74,5 @@ ENV R_RD4PDF="times,hyper"
 
 COPY package.centos-8 /package.sh
 COPY build.sh .
+COPY patches /patches
 ENTRYPOINT ./build.sh
--- a/builder/Dockerfile.debian-10
+++ b/builder/Dockerfile.debian-10
@@ -23,4 +23,5 @@ ENV PAGER /usr/bin/pager
 
 COPY package.debian-10 /package.sh
 COPY build.sh .
+COPY patches /patches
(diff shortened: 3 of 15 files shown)


builder/Dockerfile.centos-7
```diff
@@ -73,4 +73,5 @@ ENV R_RD4PDF="times,hyper"
 
 COPY package.centos-7 /package.sh
 COPY build.sh .
+COPY patches /patches
```
glin, line 76: Based on previous notes, CentOS 7 was supposedly unaffected, and I do see an installer up at https://cdn.rstudio.com/r/centos-7/pkgs/R-3.3.0-1-1.x86_64.rpm

glin (commented):
Nice! We've had this around forever and can finally remove the caveat note from the docs.

Just one more thing, handler.py will need to be updated to not skip these builds anymore: https://github.com/rstudio/r-builds/blob/b541a8ff7cb0d8a459338520cc5bec99a2076048/handler.py#L149-L154

This should also close https://github.com/rstudio/r-builds/issues/76 and https://github.com/rstudio/r-builds/issues/36

builder/Dockerfile.centos-7
gaborcsardi: Oh, OK. OTOH, it also does not hurt to do it for CentOS 7 as well, and then it is simpler if we do it everywhere.

gaborcsardi:
OK, I updated `handler.py` hopefully it is OK now.

Let me know if you want to remove the patch from CentOS 7. I think it is better to keep it, one less special case. I did try the build with the patch on CentOS 7 (in the arm64 builds) and it works fine.

glin (approved):

builder/Dockerfile.centos-7
glin: Sure, that's a fine reason to keep it in then.

PR #180, opened 2023-07-25
Add R builds for openSUSE 15.5 / SLES 15 SP5

glin:
This was a copy-paste of SUSE 15.4 with no change. The SUSE 15.4 binaries also seem to work just fine on 15.5, but better to have a separate build anyway in case SUSE makes another breaking change in the minor release.

### DIFF
--- a/builder/Dockerfile.opensuse-155
+++ b/builder/Dockerfile.opensuse-155
@@ -1,0 +1,103 @@
+FROM opensuse/leap:15.5
+
+ENV OS_IDENTIFIER opensuse-155
+
+# Most of these packages, and also the configure options that follow were determined
+# by reviewing "R-base.spec" from the following OpenSUSE RPM:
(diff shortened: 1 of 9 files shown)


builder/Dockerfile.opensuse-155
```diff
@@ -0,0 +1,103 @@
...
+    gcc \
+    gcc-c++ \
+    gcc-fortran \
+    glib2-devel \
+    glibc-locale \
+    help2man \
+    # Like openSUSE/SLES 15.3 and 15.4, SUSE 15.5 ships with both ICU 65 (libicu-devel) and
+    # ICU 69 (icu.691-devel). Prefer ICU 69 because the devel packages can't coexist, and
+    # zypper resolves libicu-devel to icu.691-devel when installing RPMs.
+    icu.691-devel \
```
glin, line 24: I verified this part manually on SUSE 15.5 - still an issue with 15.5.

PR #181, opened 2023-07-25
Add builds for Fedora 37 and 38

rundel:
This is an attempt at adding builds for the current versions of Fedora. The dockerfiles and package files are almost entirely based on the rhel-9 implementation.

I believe I have included all necessary changes as described in the readme and a test run was successful on the forked branch (https://github.com/rundel/r-builds/actions/runs/5662060100)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ R binaries are built for the following Linux operating systems:
 - Red Hat Enterprise Linux 7, 8, 9
 - openSUSE 15.4
 - SUSE Linux Enterprise 15 SP4
+- Fedora 37, 38
 
 Operating systems are supported until their vendor end-of-support dates, which
 can be found on the [Posit Platform Support](https://posit.co/about/platform-support/)
@@ -145,6 +146,25 @@ Then install the package:
 ```bash
 sudo zypper --no-gpg-checks install R-${R_VERSION}-1-1.x86_64.rpm
 ```
+
+Download the deb package:
+```bash
+# Debian 10
+curl -O https://cdn.posit.co/r/fedora-37/pkgs/r-${R_VERSION}_1_amd64.rpm
+
+# Debian 11
+curl -O https://cdn.posit.co/r/fedora-38/pkgs/r-${R_VERSION}_1_amd64.rpm
+
+# Debian 12
+curl -O https://cdn.posit.co/r/debian-12/pkgs/r-${R_VERSION}_1_amd64.deb
+```
+
+Then install the package:
+```bash
+sudo dnf install r-${R_VERSION}_1_amd64.rpm
+```
+
+
 
 ### Verify R installation
 
--- a/README.md
+++ b/README.md
@@ -147,16 +147,15 @@ Then install the package:
 sudo zypper --no-gpg-checks install R-${R_VERSION}-1-1.x86_64.rpm
 ```
 
-Download the deb package:
+#### Fedora Linux
+
+Download the rpm package:
 ```bash
-# Debian 10
+# Fedora 37
 curl -O https://cdn.posit.co/r/fedora-37/pkgs/r-${R_VERSION}_1_amd64.rpm
 
-# Debian 11
+# Fedora 38
 curl -O https://cdn.posit.co/r/fedora-38/pkgs/r-${R_VERSION}_1_amd64.rpm
-
-# Debian 12
-curl -O https://cdn.posit.co/r/debian-12/pkgs/r-${R_VERSION}_1_amd64.deb
 ```
 
 Then install the package:
--- a/builder/Dockerfile.fedora-37
+++ b/builder/Dockerfile.fedora-37
@@ -67,7 +67,8 @@ ENV CONFIGURE_OPTIONS="\
     --with-system-valgrind-headers \
     --with-tcl-config=/usr/lib64/tclConfig.sh \
     --with-tk-config=/usr/lib64/tkConfig.sh \
-    --enable-prebuilt-html"
+    --enable-prebuilt-html \
+    --with-blas=flexiblas"
 
 # Make sure that patching Java does not break R.
 # R's default JAVA_HOME path includes the exact Java version on CentOS/RHEL, which
--- a/builder/Dockerfile.fedora-37
+++ b/builder/Dockerfile.fedora-37
@@ -58,7 +58,7 @@ RUN curl -LO "https://github.com/goreleaser/nfpm/releases/download/v2.18.1/nfpm_
 
 RUN chmod 0777 /opt
 
-# Configure flags for RHEL 9 that don't use the defaults in build.sh
+# Configure flags for that don't use the defaults in build.sh
 ENV CONFIGURE_OPTIONS="\
     --enable-R-shlib \
     --with-tcltk \
--- a/builder/Dockerfile.fedora-37
+++ b/builder/Dockerfile.fedora-37
@@ -68,7 +68,8 @@ ENV CONFIGURE_OPTIONS="\
     --with-tcl-config=/usr/lib64/tclConfig.sh \
     --with-tk-config=/usr/lib64/tkConfig.sh \
     --enable-prebuilt-html \
-    --with-blas=flexiblas"
+    --with-blas=flexiblas \
+    --with-lapack=flexiblas"
 
 # Make sure that patching Java does not break R.
 # R's default JAVA_HOME path includes the exact Java version on CentOS/RHEL, which
--- a/builder/Dockerfile.fedora-37
+++ b/builder/Dockerfile.fedora-37
@@ -1,0 +1,86 @@
+FROM fedora:37
+
+ENV OS_IDENTIFIER fedora-37
+
+RUN dnf -y upgrade \
+    && dnf -y install dnf-plugins-core \
+    && dnf -y install \
+    autoconf \
+    automake \
+    bzip2-devel \
+    cairo-devel \
+    flexiblas-devel \
+    gcc-c++ \
+    gcc-gfortran \
+    java-11-openjdk-devel \
+    libICE-devel \
+    libSM-devel \
+    libX11-devel \
+    libXmu-devel \
+    libXt-devel \
+    libcurl-devel \
+    libicu-devel \
+    libjpeg-devel \
+    libpng-devel \
+    libtiff-devel \
+    libtool \
+    make \
+    ncurses-devel \
+    pango-devel \
+    pcre-devel \
+    pcre2-devel \
+    readline-devel \
+    rpm-build \
+    tcl-devel \
+    tex \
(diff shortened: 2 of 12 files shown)


glin:
Hi @rundel, thanks for the PR. So the builds here are only meant to cover the [platforms that Posit supports](https://posit.co/about/platform-support/), and we don't have Fedora there. We're talking about whether it would be useful to offer Fedora R binaries without officially supporting Fedora, and whether it's worth the additional maintenance effort. So far we're not seeing much customer/user demand and seem to have more requests for Amazon Linux (AL2023 is based on Fedora), and may go for Amazon Linux builds directly if the Fedora need is more for AL2023 support.

Could you explain a bit about your use case for Fedora support, or if you know of other communities that would benefit from Fedora builds? Is the Fedora support for running R on Amazon Linux by any chance?

And also, since we have no immediate plans to officially support Fedora, would it be feasible to either 1) fork this repo, build and upload Fedora binaries yourself, or 2) merge Fedora support here but not officially support it or publicize it in our documentation?

rundel:
Thanks for taking the time to respond and look things over. This is based on a need for my department (Statistical Science at Duke) - our use of Fedora is largely due to historical reasons (we were originally on Scientific linux which then progressed to RHEL / Centos -> Fedora). RHEL hasn't been viable for us due to the exceedingly slow rate of updates for system libraries which made managing our R package ecosystem painful (maybe this is better now w/ package manager). If I had my way we would have switch to Ubuntu awhile ago but that hasnt been a viable option due to our limited IT support and other more pressing needs.

I would be reluctant to try to support the builds myself because I don't feel like I have my head fully wrapped around the whole build process and toolchain (getting this to work was based on cribbing heavily from the existing rhel builds). I'm sure I can get it to work for our specific systems but I would certainly benefit from the fedora builds "automatically" inheriting from updates and changes that are made more generally for this repo.

I'm very much of aware of the direct and indirect costs involved with adding support for another distro and don't want to add any unnecessary extra burden on you or posit. I would be more than happy with the "unofficial" support option you've suggested if that also works for you all.

README.md
```diff
@@ -146,6 +147,25 @@ Then install the package:
...
+```bash
+# Debian 10
+curl -O https://cdn.posit.co/r/fedora-37/pkgs/r-${R_VERSION}_1_amd64.rpm
+
+# Debian 11
+curl -O https://cdn.posit.co/r/fedora-38/pkgs/r-${R_VERSION}_1_amd64.rpm
+
+# Debian 12
+curl -O https://cdn.posit.co/r/debian-12/pkgs/r-${R_VERSION}_1_amd64.deb
+```
```
glin, line 160: We'll need a new header for Fedora, and then the references to deb/Debian should be removed here

builder/Dockerfile.fedora-37
```diff
@@ -0,0 +1,86 @@
...
+# Configure flags for RHEL 9 that don't use the defaults in build.sh
+ENV CONFIGURE_OPTIONS="\
+    --enable-R-shlib \
+    --with-tcltk \
+    --enable-memory-profiling \
+    --with-x \
+    --with-system-valgrind-headers \
+    --with-tcl-config=/usr/lib64/tclConfig.sh \
+    --with-tk-config=/usr/lib64/tkConfig.sh \
+    --enable-prebuilt-html"
```
glin, line 70: For RHEL 9, we had a complicated config split in R 4.3.0 to switch to FlexiBLAS instead of OpenBLAS, because R <= 4.2 were built with OpenBLAS and we didn't want to break users.

Since these are new platforms, we can go ahead with always using FlexiBLAS since that's Fedora's preferred way of BLAS switching. So in the `CONFIGURE_OPTIONS`, we should add these same `--with-blas` and `--with-lapack` flags: https://github.com/rstudio/r-builds/blob/03dbd456d48d2e58f22cb5c8afd8b8e092709f50/builder/build.sh#L148-L154

And then in the packaging script, there are some OpenBLAS handling commands that can now be removed.

builder/Dockerfile.fedora-37
```diff
@@ -0,0 +1,86 @@
...
+
+RUN if [ "$(arch)" == "aarch64" ]; then echo arm64; else echo amd64; fi > /tmp/arch
+
+RUN curl -LO "https://github.com/goreleaser/nfpm/releases/download/v2.18.1/nfpm_$(cat /tmp/arch).rpm" && \
+    dnf install -y "nfpm_$(cat /tmp/arch).rpm" && \
+    rm "nfpm_$(cat /tmp/arch).rpm"
+
+RUN chmod 0777 /opt
+
+# Configure flags for RHEL 9 that don't use the defaults in build.sh
```
glin, line 61:
```suggestion
# Configure flags that don't use the defaults in build.sh
```

builder/Dockerfile.fedora-38
```diff
@@ -0,0 +1,86 @@
...
+RUN if [ "$(arch)" == "aarch64" ]; then echo arm64; else echo amd64; fi > /tmp/arch
+
+RUN curl -LO "https://github.com/goreleaser/nfpm/releases/download/v2.18.1/nfpm_$(cat /tmp/arch).rpm" && \
+    dnf install -y "nfpm_$(cat /tmp/arch).rpm" && \
+    rm "nfpm_$(cat /tmp/arch).rpm"
+
+RUN chmod 0777 /opt
+
+# Configure flags for RHEL 9 that don't use the defaults in build.sh
+ENV CONFIGURE_OPTIONS="\
```
glin, line 62: Same comment about FlexiBLAS for Fedora 38

builder/package.fedora-37
```diff
@@ -0,0 +1,116 @@
...
+  if [ -d ${R_INSTALL_PATH} ]; then
+    rm -r ${R_INSTALL_PATH}
+  fi
+EOF
+
+  scripts="scripts:
+  postinstall: /post-install.sh
+  postremove: /after-remove.sh
+"
+fi
```
glin, line 67: With FlexiBLAS, this whole block can be removed and replaced with just a single `blas_lib=flexiblas-devel` I think. It'll probably need some testing to make sure the packaging still works correctly.

glin (commented):
@rundel Okay, let's merge this then but only document support for it in this repo, not the official Posit docs. Unless we pick it up as a supported platform, new releases or issues with the builds probably won't get immediate attention.

This PR looks generally good though. There was only one issue I noticed with the BLAS configuration, where R is being built against the internal BLAS instead of FlexiBLAS. It's because of some RHEL 9 specific stuff that'll have to be removed. I can make these changes as well, but might not get to it right away.

README.md
rundel: This is fixed by afa8fdd

builder/package.fedora-37
rundel: This is now removed, I've kept the
```
cat <<EOF >> /after-remove.sh
  if [ -d ${R_INSTALL_PATH} ]; then
    rm -r ${R_INSTALL_PATH}
  fi
```
since this was used in both branches of the conditional - if it isn;t needed I can remove it as well.

builder/Dockerfile.fedora-38
rundel: Should be fixed and looks like flexiblas is being linked based on the ldd results in the test build

builder/package.fedora-37
glin: Based on the comment, I vaguely remember it being necessary because the packaging tool, nFPM, had a limitation where RPMs weren't completely cleaned up on uninstallation. So still good to keep I believe.

glin (approved):
Looks great, thanks for updating the quick install script as well.

glin:
I'll merge this when the CI is done, and then it might take another few hours to deploy, build, and publish all the Fedora R binaries.

glin:
These should be up on the CDN now. I tested a couple R versions briefly like:
```sh
R_VERSION=4.3.1
curl -O https://cdn.rstudio.com/r/fedora-37/pkgs/R-${R_VERSION}-1-1.x86_64.rpm
dnf install -y R-${R_VERSION}-1-1.x86_64.rpm

R_VERSION=4.3.1
curl -O https://cdn.rstudio.com/r/fedora-38/pkgs/R-${R_VERSION}-1-1.x86_64.rpm
dnf install -y R-${R_VERSION}-1-1.x86_64.rpm
```

rundel:
@glin I was just able to install these and everything looks good on our system - able to swap between the versions and everything works. Thanks for helping shepherd this through.

For some reason workbench is defaulting to 4.1.3 vs 4.3.1 when both are installed - setting `rsession-which-r` in `/etc/rstudio/rserver.conf` and/or `r-version` in `/etc/rstudio/profiles` does not seem to change this behavior. We're using launcher with a local config if that matters. 

Happy to follow up with this is community if that is a better place.

glin:
@rundel That's odd, but I know there are a ton of different ways to configure R in Workbench, e.g. described at [Configuring the Default Version of R](https://docs.posit.co/ide/server-pro/r/configuring_default.html). Maybe check the R on the PATH or if users have an IDE override? Besides that, I can't really help much though, so you could try Community. Or since you're a Workbench customer, you can also open a professional support ticket.

gaborcsardi:
IDK if you already know this, but we don't seem to have Fedora builds:
```
❯ R_VERSION=4.3.1 curl -LOf https://cdn.posit.co/r/fedora-38/pkgs/r-${R_VERSION}_1_amd64.rpm
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 403

❯ R_VERSION=3.6.1 curl -LOf https://cdn.posit.co/r/fedora-38/pkgs/r-${R_VERSION}_1_amd64.rpm
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 403

❯ R_VERSION=3.6.1 curl -LOf https://cdn.posit.co/r/fedora-37/pkgs/r-${R_VERSION}_1_amd64.rpm
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 403
```

gaborcsardi:
Ah, never mind, only a copy paste error in the README.

gaborcsardi:
Btw. https://github.com/r-hub/R/releases now also has arm64 R builds for Fedora, and Fedora support is also to r-lib/rig soon.

PR #184, opened 2023-08-16
Fix R 3.0.0 - R 3.0.3 on arm64

gaborcsardi:
Otherwise recommended packages do not compile and data sets are broken in every package, including base packages.

This came up e.g. here: https://bugs.r-project.org/show_bug.cgi?id=15491 but neither the fix, nor the commit by BDR fixes this. The issue is in the vendored TRE regex library, which got updated before R 3.1.0, so this patch is a combination of those updates:
https://github.com/wch/r-source/commit/a2635929ac962a866c67b6273440d09a3b26f0cf
https://github.com/wch/r-source/commit/3b5876791fa4c26321fc415b27f920ecc3cf1dc8

I tested this with R 3.0.0 - R 3.0.3 on Debian 12, and R 3.0.0 on Fedora 37.

### DIFF
--- a/builder/build.sh
+++ b/builder/build.sh
@@ -75,6 +75,10 @@ patch_r() {
   if [ -f "/patches/R-${1}-${OS_IDENTIFIER}.patch" ]; then
       patch -p1 < "/patches/R-${1}-${OS_IDENTIFIER}.patch"
   fi
+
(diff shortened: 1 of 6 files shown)


gaborcsardi:
The CI failures are unrelated.

glin (approved):

PR #185, opened 2023-08-16
Newer Ubuntu/Debian need libdeflate for R-devel

gaborcsardi:
R-devel now links to libdeflate, so we need to
install it, and also depend on it.

### DIFF
--- a/builder/package.debian-11
+++ b/builder/package.debian-11
@@ -36,6 +36,7 @@ depends:
 - libc6
 - libcairo2
 - libcurl4-openssl-dev
+- libdeflate0
 - libglib2.0-0
 - libgomp1
 - libicu-dev
--- a/builder/Dockerfile.debian-11
+++ b/builder/Dockerfile.debian-11
@@ -7,7 +7,7 @@ RUN set -x \
   && echo 'deb-src http://deb.debian.org/debian bullseye main' >> /etc/apt/sources.list \
   && apt-get update \
   && apt-get install -y curl gcc libcurl4-openssl-dev libicu-dev \
-     libopenblas-base libpcre2-dev make python3-pip wget \
+     libopenblas-base libpcre2-dev make python3-pip wget libdeflate-dev \
   && apt-get build-dep -y r-base
 
 RUN pip3 install awscli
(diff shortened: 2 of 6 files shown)


glin (commented):
Looks good, but a couple questions - based on the R-devel news, it sounds like libdeflate is both optional and experimental currently:
> memCompress() and memDecompress() will use the libdeflate library (https://github.com/ebiggers/libdeflate) if installed. This uses the same type of compression for type = "gzip" but is 1.5-2x faster than the system libz library on some common platforms: the speed-up may depend on the library version.
>
> There is an experimental configure option --with-libdeflate-compression which uses libdeflate rather than libz to (de)compress R objects in lazy-load databases, if the libdeflate library and headers are available. Typically tasks spend up to 5% of their time on such operations, although creating lazy-data databases is one of the exceptions.

If optional, is there some bug that causes the build to fail on the newer Ubuntu/Debian? Or did libdeflate get indirectly installed in the images, causing R to link it in? If it's the latter, could we uninstall it to fix the failures instead?

I'm wondering both if libdeflate is safe enough to use yet, and if so, whether we should add it to all distros at the same time for consistency.

And also, if feasible enough, whether we can add libdeflate as a dependency only if R uses it so it doesn't get installed with all the past versions that don't need it. I saw it in the libR linker flags so maybe something like
```sh
if /opt/R/${R_VERSION}/bin/R CMD config --ldflags | grep -q '\-ldeflate'; then
   # libdeflate was included, add it to the deb dependencies
fi
```

gaborcsardi:
Good questions, I'll investigate.

builder/package.debian-11
```diff
@@ -36,6 +36,7 @@ depends:
 - libc6
 - libcairo2
 - libcurl4-openssl-dev
+- libdeflate0
```
glin, line 39: For the test failure, looks like `-ldeflate` ends up in the libR linker flags, which we test to make sure that packages like rJava can be built. Switching this to the dev package should get it working.

gaborcsardi:
Hmmm, this is a bit weird. I don't see the error in my arm64 builds, and I also don't see it locally on x86_64. Plus, seems like our Jenkins has no trouble building R-devel on Debian 12, either.

Maybe it is an issue with the cached builder container image, and we would just need to delete the GHA cache?

In any case, no need to change anything.

Issue #202, opened 2024-02-08
Oracle Linux 9 support


tiagomlalves:
Installing R in Oracle Linux 9 fails due to missing `openblas-devel` package. This package is available in Oracle's CodeReady repository, which can be enabled by running either `sudo dnf config-manager --enable ol9_codeready_builder` or `/usr/bin/crb enable`.

The above repository is not enabled because it detects the distro as being RedHat, attempting to run `subscription-manager` which does not exist, and not enabling the required repo.

Example of `/etc/os-release` for Oracle Linux 9 shows:
```
NAME="Oracle Linux Server"
VERSION="9.1"
ID="ol"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="9.1"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Oracle Linux Server 9.1"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:oracle:linux:9:1:server"
HOME_URL="https://linux.oracle.com/"
BUG_REPORT_URL="https://github.com/oracle/oracle-linux"

ORACLE_BUGZILLA_PRODUCT="Oracle Linux 9"
ORACLE_BUGZILLA_PRODUCT_VERSION=9.1
ORACLE_SUPPORT_PRODUCT="Oracle Linux"
ORACLE_SUPPORT_PRODUCT_VERSION=9.1
```

PR #203, opened 2024-02-08
Adds Oracle Linux support

tiagomlalves:
Fixes https://github.com/rstudio/r-builds/issues/202. Adds Oracle Linux support which, in version 9, requires different logic than RedHat 9 to install repositories.

Changes validated in both Oracle Linux 8 and 9 versions.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -89,6 +89,10 @@ detect_os () {
   then
    distro="Rocky"
   fi
+  if [[ $(cat /etc/os-release | grep -e "^CPE_NAME\=*" | cut -f 2 -d '=') =~ cpe:/o:oracle:linux: ]]
+  then
+   distro="Oracle"
+  fi
   if [[ $(cat /etc/os-release | grep -e "^ID\=*" | cut -f 2 -d '=') == "debian" ]]; then
     distro="Debian"
   fi
@@ -99,7 +103,7 @@ detect_os () {
 # Returns the OS version
 detect_os_version () {
   os=$1
-  if [[ "${os}" =~ ^(RedHat|Alma|Rocky|Fedora)$ ]]; then
+  if [[ "${os}" =~ ^(RedHat|Alma|Rocky|Fedora|Oracle)$ ]]; then
     # Get the major version. /etc/redhat-release is used if /etc/os-release isn't available,
     # e.g., on CentOS/RHEL 6.
     if [[ -f /etc/os-release ]]; then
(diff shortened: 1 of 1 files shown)


glin (approved):
This looks good, thanks!

Just one question from testing it out briefly, since I'm not familiar with Oracle Linux. With a stock `oraclelinux:9` Docker image, there is no repo named `crb`, and there's no `/usr/bin/crb` utility either. So the install script's `dnf config-manager --set-enabled crb` command doesn't work in this Docker image at least.

Does this work for your Oracle Linux systems? Or did you already have CodeReady Builder enabled by any chance? In this OL9 Docker image, `dnf config-manager --enable ol9_codeready_builder` does work though.

tiagomlalves:
> Just one question from testing it out briefly, since I'm not familiar with Oracle Linux. With a stock `oraclelinux:9` Docker image, there is no repo named `crb`, and there's no `/usr/bin/crb` utility either. So the install script's `dnf config-manager --set-enabled crb` command doesn't work in this Docker image at least.

You're right - neither docker image (nor my system) has `/usr/bin/crb` utility. This has been installed when running the below from `install.sh` (see [line 389](https://github.com/rstudio/r-builds/blob/main/install.sh#L389):
```
        ${SUDO} dnf install ${yes} https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
```

> Does this work for your Oracle Linux systems? Or did you already have CodeReady Builder enabled by any chance? In this OL9 Docker image, `dnf config-manager --enable ol9_codeready_builder` does work though.

As, I'm double checking the code I see calling `dnf config-manager --set-enabled crb` 😿 ... which is great reminder that tests should have been included in this PR!!

I'll fix both and push new changes.

tiagomlalves:
I've fixed my original PR and validated changes by running the following:
```
for LINUX_VERSION in 8 9; do 
  for R_VERSION in 4.3.2 3.4.2; do
    docker run \
      -i --platform=linux/amd64 \
      -v "$(pwd)/install.sh":/install.sh \
      -e R_VERSION=${R_VERSION} -e RUN_UNATTENDED=yes \
      oraclelinux:${LINUX_VERSION} \
      sh -c "bash /install.sh ; /opt/R/\${R_VERSION}/bin/R --version"
  done
done
```

This has installed successfully R `4.3.2` and `3.4.2` in both Oracle Linux 8 and 9 versions.

glin:
Awesome, it works well now. I don't think the function naming is a big deal, but we could always rename it later if really confusing :shrug:.

PR #213, opened 2024-04-17
Add support for Ubuntu 24.04

glin:
Adds Ubuntu 24.04. It's currently in beta so this is just adding early support to unblock other things. When Ubuntu 24 is officially released, we can rerun the builds and update the documentation website.

### DIFF
--- a/builder/package.ubuntu-2404
+++ b/builder/package.ubuntu-2404
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+if [[ ! -d /tmp/output/${OS_IDENTIFIER} ]]; then
+  mkdir -p "/tmp/output/${OS_IDENTIFIER}"
+fi
+
+# R 3.x requires PCRE1. On Ubuntu 24, R 3.x also requires PCRE2 for Pango support.
+pcre_libs='- libpcre2-dev'
+if [[ "${R_VERSION}" =~ ^3 ]]; then
+  pcre_libs='- libpcre2-dev
+- libpcre3-dev'
+fi
+
+deflate_libs='# - libdeflate-dev'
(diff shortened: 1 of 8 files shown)


builder/package.ubuntu-2404
```diff
@@ -0,0 +1,77 @@
...
+- libopenblas-dev
+- libpango-1.0-0
+- libpangocairo-1.0-0
+- libpaper-utils
+${pcre_libs}
+- libpng16-16t64
+- libreadline8t64
+- libtcl8.6
+- libtiff6
+- libtirpc-dev
```
glin, line 58: This was mostly identical to Ubuntu 22 besides some library name changes and the addition of `libtirpc`, which was new to me but apparently required sometimes?

> https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Useful-libraries-and-programs
>
> An implementation of XDR is required, and the R sources contain one which is likely to suffice (although a system version may have higher performance). XDR is part of RPC and historically has been part of libc on a Unix-alike. (In principle man xdr_string should tell you which library is needed, but it often does not: on some OSes it is provided by libnsl.) However some builds61 of glibc omit or hide it with the intention that the TI-RPC library be used, in which case libtirpc (and its development version) should be installed, and its headers62 need to be on the C include path or under /usr/include/tirpc.

It's also needed to link against libR, so included the `-dev` package as a dependency.
```sh
$ /opt/R/4.3.3/bin/R CMD config --ldflags
-Wl,--export-dynamic -fopenmp -L/usr/local/lib -L/opt/R/4.3.3/lib/R/lib -lR -lpcre2-8 -llzma -lbz2 -lz -ltirpc -lrt -ldl -lm -licuuc -licui18n
```

gaborcsardi (approved):

PR #215, opened 2024-05-03
OpenBLAS support for Ubuntu 20.04 (Focal)

michaelmayer2:
Previously we had re-added OpenBLAS support to our Jammy builds following a bug fix Ubuntu release. The same fix was released for Ubuntu Focal as per https://bugs.launchpad.net/ubuntu/+source/openblas/+bug/1860601/comments/14 

This PR is reintroducing support for OpenBLAS for R builds in Ubuntu 20.04 (Focal) so that our users can benefit from additional performance as a result. 

PR has been tested successfully on R 3.4.3, 4.1.3 and 4.4.0.


glin (approved):
:+1: 

Linking https://github.com/rstudio/r-builds/issues/62 as well

glin:
This is now deployed with all the Ubuntu 20 packages rebuilt.

PR #217, opened 2024-05-12
Add Fedora 40

gaborcsardi:
Once the packages are built I'll update the README as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PLATFORMS := ubuntu-2004 ubuntu-2204 ubuntu-2404 debian-10 debian-11 debian-12 centos-7 centos-8 rhel-9 opensuse-155 fedora-38 fedora-39
+PLATFORMS := ubuntu-2004 ubuntu-2204 ubuntu-2404 debian-10 debian-11 debian-12 centos-7 centos-8 rhel-9 opensuse-155 fedora-38 fedora-39 fedora-40
 SLS_BINARY ?= ./node_modules/serverless/bin/serverless.js
 
 deps:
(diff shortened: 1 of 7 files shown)


glin (approved):

glin:
@gaborcsardi The builds are failing for a range of older R versions, which seems like R 3.3 to early 4.0.x. R 4.0.0 fails but R 4.0.5 works. I'm not sure why. The error I see in every build I've checked is about bzip2:
```sh
checking for BZ2_bzlibVersion in -lbz2... yes
checking bzlib.h usability... yes
checking bzlib.h presence... yes
checking for bzlib.h... yes
checking if bzip2 version >= 1.0.6... no
checking whether bzip2 support suffices... configure: error: bzip2 library and headers are required

$ dnf info bzip2-devel
Last metadata expiration check: 0:55:32 ago on Wed May 29 20:49:33 2024.
Installed Packages
Name         : bzip2-devel
Version      : 1.0.8
Release      : 18.fc40
Architecture : x86_64
Size         : 310 k
Source       : bzip2-1.0.8-18.fc40.src.rpm
Repository   : @System
From repo    : fedora
Summary      : Libraries and header files for apps which will use bzip2
URL          : https://sourceware.org/bzip2
License      : BSD-4-Clause
Description  : Header files and a library of bzip2 functions, for developing apps
             : which will use the library.
```
If it's some R bug that would need to be patched, then I think it's fine to leave alone if you don't need these older R versions.

gaborcsardi:
Thanks! Hopefully this fixes them: https://github.com/rstudio/r-builds/pull/228

PR #219, opened 2024-05-15
Patch R 4.0.0 - R 4.3.3 for CVE-2024-27322

gaborcsardi:
We also update the NEWS file, so by grepping for
'CVE-2024-27322' one can tell if the patched version is installed or not.

### DIFF
--- a/builder/patches/R-4.0.0.patch
+++ b/builder/patches/R-4.0.0.patch
@@ -45,18 +45,20 @@ index a389f71311..a190fbf8f3 100644
      SEXP object, icon, type, ver, fun;
      object = CAR(args); args = CDR(args);
      icon = CAR(args); args = CDR(args);
-diff --git a/doc/NEWS b/doc/NEWS
-index 1e5bc60..1ce0c7b 100644
---- a/doc/NEWS
-+++ b/doc/NEWS
-@@ -1,5 +1,10 @@
- R News
+diff --git a/doc/NEWS.Rd b/doc/NEWS.Rd
+index 094396b..75fae89 100644
+--- a/doc/NEWS.Rd
++++ b/doc/NEWS.Rd
+@@ -6,6 +6,12 @@
+ \encoding{UTF-8}
  
-+CHANGES IN POSIT'S BUILD FROM https://github.com/rstudio/r-builds
-+
-+    * readRDS() and unserialize() now signal an error instead of returning a PROMSXP,
-+      to fix CVE-2024-27322.
-+
- CHANGES IN 4.0.0:
+ \section{\Rlogo CHANGES IN 4.0.0}{
++  \subsection{CHANGES IN POSIT'S BUILD FROM \url{https://github.com/rstudio/r-builds}}{
++    \itemize{
++      \item readRDS() and unserialize() now signal an error instead of
++      returning a PROMSXP, to fix CVE-2024-27322.
++    }
++  }
  
-   SIGNIFICANT USER-VISIBLE CHANGES:
+   \subsection{SIGNIFICANT USER-VISIBLE CHANGES}{
+     \itemize{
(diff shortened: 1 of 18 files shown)


builder/patches/R-4.0.0.patch
```diff
@@ -0,0 +1,62 @@
...
+-
++    if (PRIMVAL(op) == 2) //return R_unserialize(CAR(args), CADR(args));
++	return checkNotPromise(R_unserialize(CAR(args), CADR(args)));
+     SEXP object, icon, type, ver, fun;
+     object = CAR(args); args = CDR(args);
+     icon = CAR(args); args = CDR(args);
+diff --git a/doc/NEWS b/doc/NEWS
+index 1e5bc60..1ce0c7b 100644
+--- a/doc/NEWS
++++ b/doc/NEWS
```
glin, line 51: The patch appears to be working, and I see it in the source files. But I don't see the NEWS file:
```sh
$ head /opt/R/4.0.0/lib/R/doc/NEWS
R News

CHANGES IN 4.0.0:

  SIGNIFICANT USER-VISIBLE CHANGES:

    * Packages need to be (re-)installed under this version (4.0.0) of
      R.

    * matrix objects now also inherit from class "array", so e.g.,
 ```

I actually don't see a `doc/NEWS` file in the source either though. Is it supposed to be `doc/NEWS.Rd`, and then `doc/NEWS` gets generated off that?

**Comment by stevenolen** [comment]
That `NEWS` change implies that users will need to reinstall all packages in their existing environments -- is that correct?

**Comment by gaborcsardi** [comment]
> That NEWS change implies that users will need to reinstall all packages in their existing environments -- is that correct?

No, no need to reinstall packages.

File: builder/patches/R-4.0.0.patch (hunk 1 above)
Comment (gaborcsardi) [reply builder/patches/R-4.0.0.patch:51]: Yeah, seems like they re-generate it, I am not sure why. You are right, we need to patch `NEWS.Rd`, I'll do that.

**Comment by stevenolen** [comment]
> But sensible people install packages into libraries that are specific for a certain (minor) R version, so in practice no need to reinstall anything.

Ah, yes. Definitely a standard expectation for us.

**Comment by gaborcsardi** [comment]
@glin OK, this should be better now. All the other NEWS files (`NEWS`, `NEWS.pdf`, `NEWS.html`) are re-generated at install time, so we don't patch those.

To check if your R version is patched, search for
CVE-2024-27322' in the NEWS. 

From the command line:
```sh
grep -C 5 CVE-2024-27322 /opt/R/4.0.0/lib/R/doc/NEWS
```

```
  CHANGES IN POSIT'S BUILD FROM <URL:
  https://github.com/rstudio/r-builds>:

    * readRDS() and unserialize() now signal an error instead of
      returning a PROMSXP, to fix CVE-2024-27322.

  SIGNIFICANT USER-VISIBLE CHANGES:

    * Packages need to be (re-)installed under this version (4.0.0) of
      R.
```
    
From R:
```r
options(browser = "false")
news(grepl("CVE-2024-27322", Text))
```

```
                        Changes in version 4.0.0                        

CHANGES IN POSIT'S BUILD FROM <URL:

    o   readRDS() and unserialize() now signal an error instead of
        returning a PROMSXP, to fix CVE-2024-27322.
```

(I am not sure why the URL does not show up in R, probably an R bug.)

glin (approved):
Looks good, thanks for taking care of this.

glin:
@gaborcsardi Yeah it seems like an R bug. At least with 4.3.3, the URL shows up correctly:
```r
> options(browser = "false")
> news(grepl("CVE-2024-27322", Text))

                        Changes in version 4.3.3                        

CHANGES IN POSIT'S BUILD FROM <https://github.com/rstudio/r-builds>

    o   readRDS() and unserialize() now signal an error instead of
	returning a PROMSXP, to fix CVE-2024-27322.
```

PR #221, opened 2024-05-20
Upgrade python to 3.11, upgrade packages to support python 3.11

jforest:


requirements.txt
```diff
@@ -1,14 +1,13 @@
...
 idna==2.7
 jmespath==0.9.3
 python-dateutil==2.7.5
-requests==2.20.1
-s3transfer==0.1.13
-six==1.11.0
-urllib3==1.24.2
+requests==2.31.0
+s3transfer==0.7.0
+six==1.16.0
```
jspiewak, line 12: wonder what still needs `six`

jspiewak (approved):

requirements.txt
jforest: I didn't really investigate that so much as just make sure the version didn't have any breaking changes

jspiewak: I figured, which is why I approved it.
But it was curious that we might need it at this point.

PR #234, opened 2024-07-31
Fix CentOS 7 build

glin:
Fix CentOS 7 builds - use vault.centos.org now that the mirrors are removed.

CentOS 7 is EOL but we still need it around for a little longer before fully deprecating it.

Also switch deprecated docker-compose commands to `docker compose`. This was also causing failures in GHA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ deps:
 	npm install
 
 docker-build:
-	@cd builder && docker-compose build --parallel
+	@cd builder && docker compose build --parallel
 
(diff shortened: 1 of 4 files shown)


christierney (approved):

glin:
The EPEL 7 RPM URL was taken down, although `yum install epel-release` still works. But we need the RPM URL for RHEL, so I found something that appears to work in the EPEL archive? https://dl.fedoraproject.org/pub/archive/epel/7/x86_64/Packages/e/epel-release-7-14.noarch.rpm

That should be the last thing to get this passing hopefully.

PR #252, opened 2025-04-17
Remove Debian 11 for EOL

glin:


glin:
@nodivbyzero Good catch, will fix

PR #254, opened 2025-04-24
Add Fedora 42 builds

gaborcsardi:
Fedora 42 comes with gcc 15, and it seems that all R versions, except for R-devel will require patches to compile with it.

 Seems like we need this for Tcl/Tk:
 https://github.com/wch/r-source/commit/5c5c426023c4f2c9d22582a5f94f9b3337f4973b
 https://github.com/wch/r-source/commit/5c5c426023c4f2c9d22582a5f94f9b3337f4973b.patch

Plus, we need to compile in C11 mode for R < 4.5.0.

Once we have the builds, I'll submit another PR to update the README.


glin (approved):
Looks good, I can kick off staging and then prod builds as soon as this merges.

gaborcsardi:
@glin Thanks!

glin:
@gaborcsardi Fedora 42 builds should all be up now. A few of the 3.0.x failed to build but I think that was expected from the previous Fedora versions.

PR #256, opened 2025-05-09
Migrate building and publishing to GitHub Actions; add ARM64 support

glin:
For https://github.com/rstudio/r-builds/issues/149, and also a step closer to Cloudsmith publishing / GPG signing. Also should close any other Serverless-related issues assuming this works well. We won't remove Serverless until after we can test and ensure this works well.

This adds publishing to the existing GHA test workflow, which previously only built and tested R builds:
- Add publishing to S3 staging and production. Like the current setup, pushes to `main` do not publish to S3. Publishing is only done with manual workflow triggers or the scheduled builds.
- Add scheduled GHA workflow to check for new R versions and automatically build them
- Add scheduled GHA workflow to automatically build R-devel and R-next daily

It also adds ARM64 support, which was really simple with GHA's native ARM runners. All the existing Docker images and build script already worked fine on arm64 since https://github.com/rstudio/r-builds/pull/152. DEB/RPMs are named the same way but with `aarch64`/`arm64`. Tarballs have `-arm64` appended to the end, and the x86 tarballs are left as-is to not break existing users.

After this merges:
- [ ] We'll manually trigger builds and publishes for arm64 only, staging and then production, to test the workflow in a lower risk way
- [ ] Test the two scheduled workflows
- [ ] Test the x86_64 builds
- [ ] Update the readme and docs.posit.co once the arm64 builds have been validated
- [ ] And later, remove the Serverless infra once the GHA jobs are confirmed working and stable
- [ ] There is also a SNS notification for new R builds in the old Serverless setup. I need to look into whether that's still used, and possibly replace it with a similar slack notification if so.


glin:
I've tested an actual staging publish here in this branch for lack of a better way, as the manual GHA trigger isn't usable yet. The working run is here: https://github.com/rstudio/r-builds/actions/runs/14934350944/job/41960209337. And checked that the builds and versions.json were published to S3 staging just fine.

Makefile
```diff
@@ -40,19 +40,25 @@ serverless-deploy.%: deps fetch-serverless-custom-file
 	$(SLS_BINARY) deploy --stage $* --verbose
 
 define GEN_TARGETS
+# Use PLATFORM_ARCH to override the architecture, e.g. PLATFORM_ARCH=linux/arm64 or PLATFORM_ARCH=linux/amd64
+# If unset, PLATFORM_ARCH will default to the architecture of the host machine.
 docker-build-$(platform):
-	@cd builder && docker compose build $(platform)
+	@cd builder && PLATFORM_ARCH=$(PLATFORM_ARCH) docker compose build $(platform)
```
glin, line 46: Not used in CI, only added for local testing via emulation.

manage_r_versions.py
```diff
@@ -0,0 +1,97 @@
+import argparse
```
glin, line 1: Mostly copied over from `handler.py`

.github/workflows/build.yml
```diff
@@ -45,18 +77,19 @@ jobs:
...
     strategy:
       matrix:
         platform: ${{ fromJson(needs.setup-matrix.outputs.platforms) }}
-    runs-on: ubuntu-latest
-    name: Docker image (${{ matrix.platform }})
+        arch: [amd64, arm64]
+    runs-on: ${{ matrix.arch == 'amd64' && 'ubuntu-latest' || 'ubuntu-24.04-arm' }}
+    name: Docker image (${{ matrix.platform }}-${{ matrix.arch }})
     steps:
       - uses: actions/checkout@v3
```
nodivbyzero, line 94: Should we consider using `actions/checkout@v4` instead?

nodivbyzero (approved):
👍

.github/workflows/build.yml
glin: Done

gaborcsardi (approved):
Awesome!

PR #260, opened 2025-05-23
Add Slack notification for new R builds to replace SNS notifications

glin:
Add a Slack notification upon new R version builds to replace the existing SNS notifications for Hosted. The message will be something like: `R-builds - new R versions built and published: 4.5.1`

This will need to be tested before merging.

### DIFF
--- a/.github/workflows/check-r-versions.yml
+++ b/.github/workflows/check-r-versions.yml
@@ -73,7 +73,7 @@ jobs:
   # Notify Hosted about new R versions
   notify-slack-success:
     needs: [build-new-r-versions, check-r-versions]
-    if: ${{ success() && needs.check-r-versions.outputs.new_r_versions != '' }}
+    if: ${{ needs.check-r-versions.outputs.new_r_versions != '' }}
     runs-on: ubuntu-latest
     steps:
       - name: Notify Slack on successful new R version build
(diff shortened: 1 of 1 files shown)


samperman (approved):

.github/workflows/check-r-versions.yml
```diff
@@ -69,3 +69,17 @@ jobs:
       r_versions: ${{ needs.check-r-versions.outputs.new_r_versions }}
       publish: ${{ inputs.publish || 'production' }}
     secrets: inherit
+
+  # Notify Hosted about new R versions
+  notify-slack-success:
+    needs: [build-new-r-versions, check-r-versions]
+    if: ${{ success() && needs.check-r-versions.outputs.new_r_versions != '' }}
```
jforest, line 76: I'm not sure I'm following the logic here.  If what is successful?  Does that mean if both the `needs` are successful?

glin: Right, the intent was to only notify if both `needs` succeeded. Looking again at the GHA docs, it's unclear to me if this will really work though: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions

It talks about "previous steps" succeeding but I don't know if that covers dependent jobs in `needs`. I can keep looking here to see if there's a better documented way to do this.

PR #267, opened 2025-06-26
Remove Fedora 40 for EOL

glin:
Noticed while doing the RHEL 10 update, Fedora 40 went EOL on May 13, 2025


builder/build.sh
```diff
@@ -64,14 +64,17 @@ patch_r() {
   cd /tmp/R-${1}
 
   if [ -f "/patches/R-${1}.patch" ]; then
+      echo "Applying patch /patches/R-${1}.patch"
```
glin, line 67: Added because at first, I couldn't figure out why removing the fedora-40 patches broke the fedora-41 builds

gaborcsardi:
We should probably have a better way to share patches between builds....

PR #268, opened 2025-07-08
remove gdebi from install script

edavidaja:
this tracks updates to our public documentation now that `apt` provides all the relevant desirable behavior

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -304,8 +304,7 @@ install_deb () {
   echo "Updating package indexes..."
   ${SUDO} apt-get update
   echo "Installing ${installer_name}..."
-  ${SUDO} apt-get install ${yesapt} gdebi-core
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -304,8 +304,7 @@ install_deb () {
   echo "Updating package indexes..."
   ${SUDO} apt-get update
   echo "Installing ${installer_name}..."
-  ${SUDO} apt-get install ${yesapt} gdebi-core
-  ${SUDO} gdebi ${yes} "${installer_name}"
+  ${SUDO} apt-get install ${yes} "${installer_name}"
```
glin, line 307: Looks like this is failing in tests from ending up as `apt-get install --n`. First, `--n` looks like a bug but it seems unused, could be removed anyway? And then I believe we need `$yesapt` instead here.

PR #270, opened 2025-10-29
R builds on Cloudsmith

### LINKED ISSUE #7: GPG signed binaries

jstruzik:
This PR aims to make our R builds available on Cloudsmith to allow for signed binaries and closes https://github.com/rstudio/r-builds/issues/7.

We introduce a new `cloudsmith-publish` job in the GHA workflow to optionally publish to Cloudsmith (`posit/open`). This works in two ways:

- For manual R build jobs, packages are versioned per-platform/R-version in `package-versions.json` (defaulting to `1`). To re-publish a build, bump its entry, since Cloudsmith has package overwriting disabled.
- For devel R builds, the nightly job appends a date-based version to avoid conflicts.

My main concern is how quickly the `posit/open` namespace will become populated with these builds.


README.md
```diff
@@ -354,14 +369,45 @@ The [Check for new R versions](https://github.com/rstudio/r-builds/actions/workf
...
 
 After making any changes to R-builds, this workflow may be run manually to test the changes in `staging` first. Then,
 the workflow can be rerun for `production` to build new binaries or rebuild existing binaries.
 
+### Cloudsmith Publishing
+
+R builds can be published to Cloudsmith repositories in addition to S3. To publish to Cloudsmith:
+
+1. Set `publish_cloudsmith` to your Cloudsmith repository (e.g., `posit/r-builds`)
+2. Set `package_version` for versioning (defaults to `1` for stable releases, `YYYYMMDD` for devel builds)
```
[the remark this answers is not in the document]
glin: I guess we should ask @gaborcsardi directly, would these URL changes affect rig and how? Do we need to add a fixed "latest" version URL for cdn.posit.co, or would rig prefer to add the apt/rpm/zypper repos?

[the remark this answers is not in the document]
jstruzik: > So for Cloudsmith publishing the release R versions, all R version packages always get the same package version?

That's correct. We could potentially try to add logic that increments based on the latest package version of the R build. That would require something similar where we have a pointer to what the latest "version" is or we'd have to search and sort all of the builds from Cloudsmith directly.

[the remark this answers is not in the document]
gaborcsardi: Thanks for pinging me! Great stuff! I have some questions:

1. So we'd still keep the existing CDN with fixed file names? rig does not really need that. (Well, i'd need some small updates in the web service that tells rig what to install.) But I am sure that other hardcoded the URLs. I think it would make sense to publish the latest versions with a fix URL, and then we could see how much people use them and deprecate and drop them later if not much.

2. If all builds have the same `package_version`, then for every change we'll need to rebuild all of them, right? Even if it is a small bug fix in a single build? Plus this also implies that people using the repos will get (spurious) updates, right? Additionally, if we'll also need to pay more for Cloudsmith storage (IDK how much more, though).

3. Honestly, I think we could also do this "properly" (i.e. like distros do it for their packages), and have a per-build (i.e. per RPM/DEB package) version, and include these `package_version` numbers in this repo. This way there are no spurious updates for users, we need to rebuild much less packages and there'd be less "duplicate" packages on Cloudsmith. With this we could have a proper changelog file for each package, etc. If needed mass-updates for changelog file, it is easy to script that.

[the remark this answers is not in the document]
gaborcsardi: Yes, that could definitely work. Maybe the default could always be '1', no? When we add a new distro/release, it should start with 1?

(I am sorry, I did not see your reply until now, I am not sure why. :()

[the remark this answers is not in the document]
jstruzik: @gaborcsardi no worries! (I'm late to reply as well). The [latest change](https://github.com/rstudio/r-builds/pull/270/commits/3e3b24534e232a72e9137c68053638430009c3d8) took your idea of defaulting to `1` so we don't have to have an entry for every distro in that file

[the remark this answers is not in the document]
gaborcsardi: Great!

.github/workflows/build.yml
```diff
@@ -259,3 +326,124 @@ jobs:
...
+          DRY_RUN_FLAG=""
+          if [[ "${{ inputs.cloudsmith_dry_run }}" == "true" ]]; then
+            echo "Running in dry-run mode"
+            DRY_RUN_FLAG="--dry-run"
+          fi
+
+          echo "Publishing $pkg_file to ${{ inputs.publish_cloudsmith }}/${{ steps.pkg-info.outputs.distro }}"
+          
+          cloudsmith push ${{ steps.pkg-info.outputs.pkg_type }} \
+            --republish \
```
glin: Is --republish doing anything if we're not allowed to republish on cloudsmith?

.github/workflows/build.yml
```diff
@@ -208,22 +251,46 @@ jobs:
             --load \
             builder
 
+      - name: Resolve package version
+        id: pkg-version
+        run: |
+          if [[ -n "${{ inputs.package_version }}" ]]; then
+            echo "version=${{ inputs.package_version }}" >> $GITHUB_OUTPUT
```
glin: I think this is low risk and not a big deal, but a claude review flags this as theoretically possible shell injection if a repo fork were to run this workflow (which it shouldn't be able to). But it's also fairly straightforward to fix by passing the inputs as environment variables instead:

> Since you control the repo and all `workflow_call` callers are internal (`devel-daily.yml`), there's no realistic attack vector today. The risk would only materialize if:
> 
> - An external repo calls your reusable workflow via `workflow_call`
> - A fork submits a PR that somehow triggers the workflow (not applicable here since the inputs are manual/scheduled)
> 
> It's still considered a best practice to avoid direct `${{ }}` interpolation in `run:` blocks, because it's a latent vulnerability that becomes real if usage patterns change. But it's not urgent.
> 
> **Fix example** - use `env:` to pass values as environment variables instead of string-interpolating them into shell:
> 
> ```yaml
> # Before (vulnerable pattern):
> - name: Resolve package version
>   id: pkg-version
>   run: |
>     if [[ -n "${{ inputs.package_version }}" ]]; then
>       echo "version=${{ inputs.package_version }}" >> $GITHUB_OUTPUT
>     else
>       version=$(jq -r '."${{ matrix.platform }}"."${{ matrix.r_version }}" // 1' package-versions.json)
>       echo "version=$version" >> $GITHUB_OUTPUT
>     fi
> 
> # After (safe pattern):
> - name: Resolve package version
>   id: pkg-version
>   env:
>     PKG_VERSION_INPUT: ${{ inputs.package_version }}
>     PLATFORM: ${{ matrix.platform }}
>     R_VERSION: ${{ matrix.r_version }}
>   run: |
>     if [[ -n "$PKG_VERSION_INPUT" ]]; then
>       echo "version=$PKG_VERSION_INPUT" >> $GITHUB_OUTPUT
>     else
>       version=$(jq -r ".[\"$PLATFORM\"][\"$R_VERSION\"] // 1" package-versions.json)
>       echo "version=$version" >> $GITHUB_OUTPUT
>     fi
> ```

.github/workflows/build.yml
```diff
@@ -259,3 +326,124 @@ jobs:
...
+            ubuntu-2004)
+              echo "distro=ubuntu/focal" >> $GITHUB_OUTPUT
+              echo "pkg_type=deb" >> $GITHUB_OUTPUT
+              echo "pkg_pattern=r-*.deb" >> $GITHUB_OUTPUT
+              ;;
+            ubuntu-2204)
+              echo "distro=ubuntu/jammy" >> $GITHUB_OUTPUT
+              echo "pkg_type=deb" >> $GITHUB_OUTPUT
+              echo "pkg_pattern=r-*.deb" >> $GITHUB_OUTPUT
+              ;;
```
glin: Since this is all manually specified, it would be good to add it to the `Adding a new platform` checklist in the README so we don't forget to update it.

Or if possible, I would love to co-locate this list with some other file that has platform info for easier maintenance. I'm not sure which though, but we have similar platform variables/scripts scattered throughout the Makefile, `get_matrix.py` script, dockerfiles, docker-compose files, quick-install script. I could think of ideas later.

.github/workflows/devel-daily.yml
```diff
@@ -4,15 +4,46 @@ on:
   schedule:
     - cron: '0 4 * * *'
   workflow_dispatch:
+    inputs:
+      publish_cloudsmith:
+        description: |
+          Publish to Cloudsmith? Specify repository (e.g., "posit/r-builds") or leave empty to skip.
+        required: false
+        default: 'posit/r-builds'
```
glin: What's the difference between `posit/r-builds` and `posit/open/r-builds`? Should this be `posit/open/r-builds`?

README.md
```diff
@@ -360,14 +375,60 @@ The [Check for new R versions](https://github.com/rstudio/r-builds/actions/workf
...
+Package versions for stable releases are managed in `package-versions.json`, keyed by
+platform and R version. Builds without an entry default to version `1`. Only add entries
+when a version bump is needed (e.g., a packaging fix for a specific R version or platform).
+
+```json
+// Example: bump centos-8 R 4.2.3 to version 2 after a targeted fix
+{
+  "centos-8": {
+    "4.2.3": 2
+  }
```
glin, line 700: Since this is manually specified, I could see it as really tedious to enumerate, especially if you're unfamiliar with this repo. I would love to see a fully automatic workflow where the version auto-bumps if any code has changed, but don't know how we'd achieve that. So manual seems fine to start out with for now.

gaborcsardi (approved):
Thanks a lot for your work on this!

README.md
glin: Apparently there is a cloudsmith API and CLI command to list the latest package(s) and what version they're at. So it seems possible for a workflow to dynamically query the latest version and auto-bump it without having to track the current version in the repo.

Here is sample code that claude came up with, totally untested but just as an example. I think we could still start out with the manual versions file, then see if an automated version bumper could work later on.

> ```yaml
> - name: Resolve package version
>   env:
>     PLATFORM: ${{ matrix.platform }}
>     R_VERSION: ${{ matrix.r_version }}
>     CS_REPO: ${{ inputs.publish_cloudsmith }}
>     PKG_VERSION_INPUT: ${{ inputs.package_version }}
>   run: |
>     if [[ -n "$PKG_VERSION_INPUT" ]]; then
>       echo "version=$PKG_VERSION_INPUT" >> $GITHUB_OUTPUT
>       exit 0
>     fi
> 
>     # Map platform to package name prefix and distro
>     # (reuse the distro mapping from the publish step)
>     case "$PLATFORM" in
>       ubuntu-*|debian-*) pkg_name="r-${R_VERSION}" ;;
>       *)                 pkg_name="R-${R_VERSION}" ;;
>     esac
>     distro="${{ steps.pkg-info.outputs.distro }}"
> 
>     # Query Cloudsmith for the current highest version
>     current=$(curl -sf \
>       "https://api.cloudsmith.io/v1/packages/${CS_REPO}/?query=name:^${pkg_name}\$ AND distribution:${distro}&sort=-version&page_size=1" \
>       | jq -r '.[0].version // "0"')
> 
>     next=$((current + 1))
>     echo "Current version: $current, next: $next"
>     echo "version=$next" >> $GITHUB_OUTPUT
> ```

.github/workflows/build.yml
jstruzik: You're right, removed

.github/workflows/devel-daily.yml
jstruzik: I changed it to just `posit/open`

glin (approved):
Looks good to merge and try out

Issue #197, opened 2024-01-22
Amazon Linux 2023 support


mmuurr:
Are there any plans to add Amazon Linux 2023 to the list of Linux distros?
Most AWS workloads are themselves shifting to this distro but AWS's software repo lags a bit, so it'd be nice to have new R versions for AL2023.
The latest pre-built R version distributed by AWS is R 4.1.2; this RPM includes the Makeconf used to produce the binary, which can be used as a template for the build scripts here, if needed.
In principle it should be very close to the Fedora 37 & 38 builds (as AL2023 inherits quite a bit from Fedora 34+).

I'm mostly just curious, as I'll likely just build R from source anyhow on my AL2023 AMIs.

gaborcsardi:
Did you try to install the Fedora 37 or Fedora 38 build on AL2023?

mmuurr:
## Testing on AL2023

Some testing on AL2023 (which [borrows from Fedora 34, 35, 36, _and_ CentOS 9 Streams](https://docs.aws.amazon.com/linux/al2023/ug/relationship-to-fedora.html)!):

Both the Fedora 37 & 38 builds fail to run (with both emitting the error: `error while loading shared libraries: libicuuc.so.71: cannot open shared object file: No such file or directory`).

The RHEL 9 build seems to work -- at least in my light testing.
* From a fresh EC2 instance with the newest AL2023 AMI (al2023-ami-2023.3.20240122.0-kernel-6.1-x86_64).
* Installed `curl` and `git` (only) to bootstrap the process.
* Fetched from `https://cdn.posit.co/r/rhel-9/pkgs/R-4.3.2-1-1.x86_64.rpm`
* Used `dnf` to install from that RPM, let `dnf` resolve dependencies. (Full output of that step included below.)
* Run the `test-r.sh` script; which as far as I can tell (which may not be very far) seemed to execute normally. The full output (if you'd like to inspect) is below.

<details>
<summary>Output of dnf install step:</summary>

```
[ec2-user@myhost]$ sudo dnf install R-4.3.2-1-1.x86_64.rpm
Last metadata expiration check: 0:11:05 ago on Tue Jan 23 07:20:46 2024.
Dependencies resolved.
===================================================================================================================================================================================
 Package                                             Architecture                   Version                                             Repository                            Size
===================================================================================================================================================================================
Installing:
 R-4.3.2                                             x86_64                         1-1                                                 @commandline                          64 M
Installing dependencies:
[...]

Issue #94, opened 2021-08-26
Feature Request: Support for custom install location


lorenzwalthert:
From the README, I see that R will be installed in `/opt/R/${R_VERSION}`. I'd like to install R in another place. Ideally, I'd like to set an env variable for this, similar to `R_VERSION`, so that the install location would be `${R_HOME}` (I think that's the variable also used in other contexts for the same meaning). Would you consider supporting this or do you know another way of installing R into a dedicated directory without installing from source?

lorenzwalthert:
Ok. Thanks. Yes, I am aware I can use that prefix, but I want to avoid build time dependencies or remove them cleanly after install, so I thought this repo can help me with that. But maybe I miss understood the purpose of this repo then.

PR #280, opened 2026-04-14
Portable + relocatable manylinux R builds

glin:
**Experimental:** I'm proposing we add this but do not widely advertise or document on https://docs.posit.co just yet. I've tried to test it thoroughly, but it definitely needs more testing to find uncovered edge cases.

---

Add support for building portable, cross-distro Linux builds of R using the same manylinux approach we've done for the portable manylinux R binary packages in Posit Package Manager.

- The binaries are based on RHEL 9 or `manylinux_2_34`, so they work for any glibc 2.34+ based distro, including all Posit-supported distros except RHEL 8. This includes support for Amazon Linux 2023 (https://github.com/rstudio/r-builds/issues/197), SLES 16 (https://github.com/rstudio/r-builds/issues/278), and Arch Linux (my personal distro where I haven't been able to use the R builds :smiling_face_with_tear:)

- They are also relocatable as a secondary benefit, which means you can install R to any arbitrary location (https://github.com/rstudio/r-builds/issues/94) as a non-root user. You just download the tarball, extract it, and R works.

A follow-up PR adds musllinux support for Alpine Linux: https://github.com/rstudio/r-builds/pull/281. I've broken that out into another PR for easier reviewing. It wasn't an original goal, but the manylinux approach worked well enough that I could get musllinux working pretty easily.

---

For more detail on how it works, see the:
- Main `README.md` with a general summary and installation steps: https://github.com/rstudio/r-builds/blob/5b1df39e86d1afea9f4ce27be904e1501ca8ffb1/README.md#portable-manylinux---any-linux-distro-with-glibc--234
- Portable R `README.md` with more technical details: https://github.com/rstudio/r-builds/blob/5b1df39e86d1afea9f4ce27be904e1501ca8ffb1/builder/portable-r/README.md
  - Covers the limitations, architecture, gotchas, and instructions for how to add a new manylinux platform
- [`make-relocatable.sh`](https://github.com/rstudio/r-builds/blob/5b1df39e86d1afea9f4ce27be904e1501ca8ffb1/builder/portable-r/make-relocatable.sh) script that patches R to be relocatable. This could potentially be applied to the regular R builds to make them relocatable without have to do a custom recompile of R.

---

I'll also extract the most important parts of the README here:

### Installation

You can install these via either generic DEB/RPM/APK packages, e.g.:
```bash
R_VERSION=4.5.3
curl -O https://cdn.posit.co/r/manylinux_2_34/pkgs/r-${R_VERSION}_1_$(dpkg --print-architecture).deb
sudo apt-get install -y ./r-${R_VERSION}_1_$(dpkg --print-architecture).deb
```

Or by downloading and extracting the tarball (does not require root):
```bash
R_VERSION=4.5.3
curl -O https://cdn.posit.co/r/manylinux_2_34/R-${R_VERSION}-manylinux_2_34.tar.gz

mkdir -p ~/R
tar xzf R-${R_VERSION}-manylinux_2_34*.tar.gz -C ~/R
~/R/${R_VERSION}/bin/R --version
```

### Limitations

These portable R builds still have two runtime dependencies that I decided to leave separate:
- CA certificates (`ca-certificates` package on most distros) are needed for HTTPS / `install.packages` support. I left them external since they're already installed on most systems, except maybe a super bare docker image. I think it's possible to bundle them (and https://github.com/r-hub/r-glibc might be doing that), but it's also nice to get CA cert updates from your distro.
- `fontconfig` config files + system fonts are needed for font rendering and plotting. Also commonly present and R starts fine without it.

In addition to tarballs, we also distribute DEB, RPM, and apk packages here with `ca-certificates` and `fontconfig` dependencies for convenience, since those package names are identical on most (or all) distros.

### Why not `manylinux_2_28`

`manylinux_2_28` (based on RHEL 8) has old system libs like OpenSSL 1.x, and had an issue with its old version of fontconfig 2.13 causing warnings under newer distros (2.13 doesn't understand modern config directives like `<reset-dirs/>`).

Since this is experimental, I figured it'd make sense to start at `manylinux_2_34` which only excludes RHEL 8. If really needed, I left the original `manylinux_2_28` setup in git history here so we can restore it.

### BLAS/LAPACK

We bundle OpenBLAS here like all other builds, and it's not swappable. We use OpenBLAS OpenMP for manylinux_2_34, matching RHEL 9. And OpenBLAS pthreads for musllinux since Alpine doesn't have an OpenMP build.

### How system libraries are bundled

See [`delocate_r.py`](https://github.com/rstudio/r-builds/blob/5b1df39e86d1afea9f4ce27be904e1501ca8ffb1/builder/portable-r/delocate_r.py), a short Python script that adapts auditwheel and auditwheel-r specifically for R, rather than for Python wheels or R packages. It is pretty much a patchelf wrapper that bundles R's shared library dependencies, rewrites rpaths with `$ORIGIN` pointing to the bundled library location, and mangles SONAMEs to prevent most types of shared library conflicts.

It's a different approach from other projects like https://github.com/r-hub/r-glibc or astral's standalone python which do static linking. 

The bundling method is probably the biggest risk here, but so far we haven't seen any issues with the manylinux binary packages, so I hope it works out here too. This approach does make it really straightforward to add new platforms since `manylinux_2_34` was almost entirely a copy of the existing RHEL 9 build.

Claude wrote this and modified it for musllinux pretty easily, so I have no concerns about it being a hand-rolled script. `auditwheel-r` reuses `auditwheel`, and I've found `auditwheel-r` difficult to update because `auditwheel` has a ton of python-specific stuff that we have to intentionally work around and ignore.

### Security

Like the manylinux binary packages, R on Windows/macOS, and python-build-standalone, there are security implications with bundling system libs like OpenSSL and libcurl that are now harder to update. It's already an issue with R on Windows/macOS and uv-installed Python, so this isn't anything too surprising.

But here, we're now including a [CycloneDX](https://cyclonedx.org/) 1.5 SBOM at `$R_HOME/sbom.cdx.json` that lists all bundled shared libraries and their source packages. This can be used to audit bundled library versions and check for known vulnerabilities.

```sh
$ cat /opt/R/4.5.3/lib/R/sbom.cdx.json
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  ...,
    "properties": [
      {
        "name": "posit:os-identifier",
        "value": "manylinux_2_34"
      },
      {
        "name": "posit:base-image",
        "value": "Rocky Linux release 9.7 (Blue Onyx)"
      }
    ]
  },
  "components": [
    "...",
    {
      "type": "library",
      "name": "libcurl-minimal",
      "version": "7.76.1-35.el9_7.3",
      "purl": "pkg:rpm/rocky/libcurl-minimal@7.76.1-35.el9_7.3?arch=x86_64",
      ...
    },
    {
      "type": "library",
      "name": "openssl-libs",
      "version": "3.5.1-7.el9_7",
      "purl": "pkg:rpm/rocky/openssl-libs@3.5.1-7.el9_7?arch=x86_64",
      "properties": [
        {
          "name": "posit:bundled-files",
          "value": "libcrypto-7a1514b3.so.3, libssl-6db2b093.so.3"
        }
      ]
    },
    ...
  ]
}
```


CLAUDE.md
```diff
@@ -0,0 +1,47 @@
+# CLAUDE.md
```
nodivbyzero, line 1: nit: This is useful but should probably be in a separate PR to keep this one focused on portable builds ???

nodivbyzero (approved):

.github/workflows/build.yml
```diff
@@ -131,6 +131,23 @@ jobs:
...
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - name: Install pytest
+        run: pip install pytest
+
+      - name: Run unit tests
+        run: make unit-test
```
jmwoliver, line 149: Could potentially use uv and the ///script syntax in the new python scripts and make target. Not a big deal if that is not the pattern in this repo though.

builder/portable-r/generate_sbom.py
```diff
@@ -0,0 +1,200 @@
...
+            for line in f:
+                if line.startswith("ID="):
+                    distro = line.strip().split("=")[1].strip('"')
+                    break
+    except FileNotFoundError:
+        pass
+    return f"pkg:rpm/{distro}/{name}@{version}?arch={arch}"
+
+
+def purl_for_apk(name: str, version: str) -> str:
```
jmwoliver, line 95: Not a big deal but there seems to be some musllinux code in this PR. It may need to be removed if we don't do #281, but it is fine for now.

builder/portable-r/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+| Dependency | Why | Package |
+|---|---|---|
+| glibc >= build version (manylinux) or musl >= 1.2 (musllinux) | Core C library | (always present) |
+| `ca-certificates` | SSL/TLS certificate bundle for HTTPS | `ca-certificates` |
+| `fontconfig` | Font config files for text rendering | `fontconfig` (on openSUSE/SLES, also install `dejavu-fonts`) |
+| `ttf-dejavu` (Alpine only) | Fonts for text rendering in plots | `ttf-dejavu` (not pulled in by fontconfig on Alpine) |
+| `which` | Required by R's `utils` package at startup | `which` (RHEL/SUSE; included in `debianutils` on Debian/Ubuntu; in `coreutils` on Alpine) |
+
+For compiling R packages from source, users also need a C/C++/Fortran compiler and dev packages for `pcre2`, `lzma`, `bz2`, `zlib`, `icu`.
```
jmwoliver, line 203: I saw your comment about this:
> These portable R builds still have two runtime dependencies that I decided to leave separate:
> - CA certificates (ca-certificates package on most distros) are needed for HTTPS / install.packages support. I left them external since they're already installed on most systems, except maybe a super bare docker image. I think it's possible to bundle them (and https://github.com/r-hub/r-glibc might be doing that), but it's also nice to get CA cert updates from your distro.
> - fontconfig config files + system fonts are needed for font rendering and plotting. Also commonly present and R starts fine without it.

Could you see adding these just to have everything all in one? Or even a different variant that includes them? I'm okay with whatever you think is best, especially since these are included in most distros.

jmwoliver (commented):
Looks good to me! A few thoughts, but we'll probably just have to get these out in the wild to be able to test them more thoroughly.

CLAUDE.md
glin: That's fair. I added it here because it was generated off the conversation and gotchas hit during this work. It could go either way, separate PR or here, but I'll just leave it here since it was a small add anyway.

.github/workflows/build.yml
glin: Oh yeah, that could work just as well. The CI workflow already used a generic `setup-python`, so that's mostly why. There also aren't any special dependencies here. But uv is worth keeping in mind for some future update.

builder/portable-r/generate_sbom.py
glin: I'll remove it if we decide not to add musllinux, but I do want to at least try it to start. And yeah, there's some musllinux stuff that was just too tricky to separate out for the PR split. The original branch had manylinux/musllinux really tangled up (https://github.com/rstudio/r-builds/commits/manylinux-r/)

builder/portable-r/README.md
glin: I want to bet on this for now and we can make changes later if needed. Fonts are probably the bigger issue, but I'm not sure how to solve that and never investigated. Gabor did have some ideas for bundling a set of minimum system fonts though.

PR #281, opened 2026-04-14
Add portable musllinux build (Alpine Linux / musl distros)

glin:
Follow-up to https://github.com/rstudio/r-builds/pull/280, adding support for `musllinux_1_2` (Alpine Linux and other musl-based distros).

See the main PR at https://github.com/rstudio/r-builds/pull/280. This is mostly the same, just split out for easier reviewing, or in case we don't actually want to support musllinux.


glin:
One last fix from testing this on Arch (https://github.com/rstudio/r-builds/pull/281/changes/bb59118a2a79dba3a4ba5300caf5cabaf1f3cf6d), but otherwise I think this is ready. I've already built `manylinux_2_34`/`musllinux_1_2`  binaries as well, and they all build fine down to R 3.0.0: https://github.com/rstudio/r-builds/actions/runs/24682635031

glin:
Also, I know there's Cloudsmith support coming up (https://github.com/rstudio/r-builds/pull/270) but I think it makes sense to leave the portable DEB/RPM/APKs out of that for now. The tarballs might be the most common installation method anyway.

builder/portable-r/test-musllinux.sh
```diff
@@ -0,0 +1,196 @@
+#!/bin/sh
```
nodivbyzero, line 1: nit: the script header says it uses busybox ash, but then calls bash for test-r.sh.
This is fine since bash is installed, but the shebang could be #!/bin/bash for consistency if bash is always available.

nodivbyzero (approved):

jmwoliver (approved):
Looks great! musllinux/alpine mostly enables a lighter weight container image right? Do you know how much smaller the musllinux build is compared to manylinux? Do you think it is worth having this support burden for the increased build matrix?

builder/portable-r/test-musllinux.sh
glin: The script installs bash as part of the test and assumes bash isn't available when first run. The alpine images don't have bash installed by default, so if we switched the shebang to `/bin/bash`, I think the script might fail to run unless we installed bash as part of the test, e.g. `command: /bin/sh -c 'apk add bash && /r-builds/builder/portable-r/test-musllinux.sh'` or something. 

Soo I'll just leave it as-is for the simplicity of the script installing its own dependencies

glin:
@nodivbyzero @jmwoliver Thanks for reviewing! And yes, we've had several requests for Alpine support in PPM throughout the years for the minimal container use case. https://github.com/rstudio/r-system-requirements/issues/148 already supports Alpine, and there are scattered requests at https://github.com/rstudio/r-docker/issues/47 and I forget where else.

Alpine releases pretty often, twice a year, so adding builds for individual releases would be annoying to maintain. So a musllinux would be much easier, maybe just a version bump every now and then so we're not pinned to an EOL alpine for too long.

For sizes, the R builds are comparable, and the main gain would be in the container image sizes. Alpine images are like <10 MB so this R build is really pushing it. For maximum minimalness, https://github.com/r-hub/r-minimal is a better choice, but I wouldn't say minimalness is a goal for these R builds.
```sh
$ du -sh builder/integration/tmp/r/musllinux_1_2/R-4.5.3-musllinux_1_2.tar.gz
91M     builder/integration/tmp/r/musllinux_1_2/R-4.5.3-musllinux_1_2.tar.gz

$ du -sh builder/integration/tmp/r/manylinux_2_34/R-4.5.3-manylinux_2_34.tar.gz
105M    builder/integration/tmp/r/manylinux_2_34/R-4.5.3-manylinux_2_34.tar.gz
```

PR #282, opened 2026-04-15
Add support for Ubuntu 26.04

nodivbyzero:
## Summary                                                                                                                                                          
                                                            
  - Add R build support for Ubuntu 26.04 (codename "Resolute Raccoon")
  - Ubuntu 26.04 does not include PCRE1 (libpcre3), so R 3.x versions are not supported on this platform (same as RHEL 10 and Debian 13)
  - Uses nfpm v2.46.1 for packaging (updated from v2.18.1 due to changed release asset naming)


glin (approved):
👍

PR #284, opened 2026-04-22
feat: portable, relocatable macOS and Windows R builds

jmwoliver:
## Overview

This PR is an **experimental**. It adds portable, relocatable macOS (arm64 + x86_64) and Windows (x86_64) R builds to r-builds, following the same goals as the Linux portable builds in #280 and #281, but using a fundamentally different approach: post-processing official CRAN binaries rather than compiling from source.

---

## Why not compile from source (like Linux)?

For Linux, we compile from source because CRAN doesn't ship Linux binaries. macOS and Windows are different: CRAN ships high-quality official binaries built with Apple's blessed toolchain (custom gfortran, correct configure flags, bundled Tcl/Tk and gfortran runtime). Replicating that toolchain from scratch would be complex and fragile. Instead, we download the CRAN binary, extract it without installing, and run a post-processing pipeline to strip hardcoded paths.

---

### macOS

CRAN `.pkg` files are Xar archives containing an Inno-style payload. We extract them with `pkgutil --expand-full` (no admin rights, no system installation) and then run a three-stage pipeline:

1. **`patch-mach-o.sh`** — Uses `install_name_tool` to rewrite all hardcoded `/Library/Frameworks/R.framework/...` load commands in every `.dylib` and `.so` to `@rpath`/`@loader_path` references, the macOS equivalent of patchelf + `$ORIGIN`. Signs all Mach-O binaries (ad-hoc always; Developer ID + notarization on production).

2. **`make-relocatable.sh`** — Patches `bin/R` and `bin/Rscript` to derive `R_HOME` at runtime from the script's own location (BSD `sed -i ''` syntax). Patches `etc/Makeconf` to replace `-framework R` with `-lR` so package compilation works from any path. Patches `R_SHARE_DIR`, `R_INCLUDE_DIR`, `R_DOC_DIR` to be `${R_HOME}`-relative.

3. **`install-rprofile-hook.sh`** — Writes `Rprofile.site` to configure a portable `site-library` inside the R directory and set PPM as the default CRAN repository.

Output: `R-{version}-macos-{arch}.tar.gz` — extract anywhere, run `bin/R`.

### Windows

Windows CRAN R is already largely self-contained (all DLLs bundled, paths largely relative). No DLL patching or binary rewriting is needed.

1. Download the CRAN `.exe` installer
2. Extract with `innoextract` (no silent install, no admin rights, no registry changes); falls back to `/VERYSILENT /CURRENTUSER` if innoextract is unavailable
3. Clean up `unins*` artifacts
4. Write `Rprofile.site` with portable `site-library` and PPM default

Output: `R-{version}-windows.zip` — extract anywhere, run `bin\R.exe`.

---

## What gets built

| Platform | Arch | R versions | Notes |
|---|---|---|---|
| macOS | arm64 | R 4.1+, last-5 + devel | No CRAN arm64 `.pkg` before R 4.1 |
| macOS | x86_64 | R 4.1+, last-5 + devel | No binaries before R 4.1 |
| Windows | x86_64 | R 3.6.3+, last-5 + 3.6.3 + devel | |

R 3.6.3 is included as a long-term compatibility target (matching existing Linux builds). For macOS, 3.6.3 is x86_64 only.

---

## Published artifacts

Following the existing CDN pattern:

```
cdn.posit.co/r/macos-arm64/R-{version}-macos-arm64.tar.gz
cdn.posit.co/r/macos-x86_64/R-{version}-macos-x86_64.tar.gz
cdn.posit.co/r/windows/R-{version}-windows.zip
```

---

## CI integration

Two new reusable workflows (`build-macos.yml`, `build-windows.yml`) follow the same OIDC auth, `manage_r_versions.py`, and S3 publish patterns as the Linux workflows:

- **`build.yml`** — `build-macos` and `build-windows` call jobs are added, gated on `contains(inputs.platforms, 'macos')`/`'windows'`
- **`check-r-versions.yml`** — automatically triggers macOS and Windows builds alongside Linux when new R versions are detected; Slack notification waits for all three platform groups

macOS CI uses native runners (`macos-14` for arm64, `macos-13` for x86_64) — no Docker involved.

---

## Testing locally

Build:
```bash
# macOS arm64 (requires Xcode Command Line Tools)
bash macos/build.sh 4.4.1 arm64 output
tar xzf output/R-4.4.1-macos-arm64.tar.gz -C output/
bash macos/test.sh output/R-4.4.1

# or via Make
R_VERSION=4.4.1 ARCH=arm64 make build-r-macos
```

Test:
```
tar xzf output/R-4.4.1-macos-arm64.tar.gz -C output/
bash macos/test.sh output/R-4.4.1
# or via Make (after extracting):
R_VERSION=4.4.1 ARCH=arm64 make test-r-macos
```

Quick smoke test without full test suite:
```
tar xzf output/R-4.4.1-macos-arm64.tar.gz -C output/
output/R-4.4.1/bin/R --version
output/R-4.4.1/bin/Rscript -e "cat(R.home(), '\n')"
```

The test suite checks: R startup, `R.home()` accuracy, Rscript, `capabilities()` (cairo/png/tcltk), BLAS/LAPACK, absence of hardcoded `/Library/Frameworks` paths in all Mach-O binaries, and relocatability (copy to new path, verify R still works).

---

## Open questions / known gaps

- **Apple notarization secrets** — the workflow supports Developer ID signing + notarization but the repo doesn't yet have the required secrets (`MACOS_DEVELOPER_CERTIFICATE`, `APPLE_TEAM_ID`, etc.). Without them, builds use ad-hoc signing (functional locally, but Gatekeeper will quarantine on end-user systems without stapling).
- **`--acl public-read`** on S3 — carried from existing publish patterns; may need adjustment if the bucket has `BucketOwnerEnforced` ownership.
- **Windows arm64** — CRAN Windows arm64 support is too new and untested; not included.
- **R devel on macOS** — URL pattern for nightly builds on `mac.r-project.org` is in place but hasn't been tested against a live devel build.
- **No manylinux/musllinux-style library bundling on macOS** — CRAN bundles all required dylibs; we rely on that remaining stable.

---

## Files changed

| Path | Description |
|---|---|
| `macos/build.sh` | Main macOS build orchestrator |
| `macos/patch-mach-o.sh` | Mach-O load command rewriting + codesigning |
| `macos/make-relocatable.sh` | bin/R, Rscript, Makeconf patching (BSD sed) |
| `macos/install-rprofile-hook.sh` | Rprofile.site configuration |
| `macos/notarize.sh` | Apple Notary Service submission |
| `macos/entitlements.plist` | Hardened runtime entitlements |
| `macos/test.sh` | Integration test suite |
| `windows/build.ps1` | Windows build orchestrator |
| `windows/test.ps1` | Windows integration tests |
| `.github/workflows/build-macos.yml` | Reusable macOS CI workflow |
| `.github/workflows/build-windows.yml` | Reusable Windows CI workflow |
| `.github/workflows/build.yml` | Extended with macOS/Windows call jobs |
| `.github/workflows/check-r-versions.yml` | Extended with macOS/Windows auto-triggers |
| `Makefile` | Local build/test targets for macOS and Windows |
| `README.md` | Installation instructions for macOS and Windows |
| `CLAUDE.md` | Developer notes |

---

## Existing art for reference

https://github.com/portable-r/portable-r-macos
https://github.com/portable-r/portable-r-windows

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,8 +43,10 @@ See [builder/patches/README.md](builder/patches/README.md).
 
 - macOS build scripts live in `macos/`, Windows scripts in `windows/`
 - These use CRAN binaries + post-processing (not compiled from source like Linux builds)
+- See [`macos/README.md`](macos/README.md) and [`windows/README.md`](windows/README.md) for technical details (Mach-O patching pipeline, code signing, Inno Setup extraction, IDE compatibility, troubleshooting)
 - CI workflows are `.github/workflows/build-macos.yml` and `.github/workflows/build-windows.yml`
 - Makefile targets: `build-r-macos`, `test-r-macos`, `build-r-windows`, `test-r-windows`
+- Portable-R startup hooks live in the **base Rprofile** (`library/base/R/Rprofile`), not `etc/Rprofile.site`, so they survive `R --vanilla` and IDE-embedded R sessions. This matches the manylinux PR #280 pattern.
 
 ## Code style
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -39,6 +39,13 @@ See [builder/patches/README.md](builder/patches/README.md).
 - **OpenBLAS naming inconsistency**: OpenBLAS .so names differ across distros (`libopenblaso.so` on Rocky 9 vs `libopenblas.so` on Ubuntu/Alpine). This is why portable builds use a post-build BLAS swap (keeping Makeconf's `-lRblas`) rather than `--with-blas=<lib>` at configure time.
 - **PCRE2 hidden for R 3.x**: `build.sh` temporarily hides PCRE2 pkg-config during configure for R < 4.0 to prevent unwanted linkage.
 
+## macOS and Windows builds
+
+- macOS build scripts live in `macos/`, Windows scripts in `windows/`
+- These use CRAN binaries + post-processing (not compiled from source like Linux builds)
+- CI workflows are `.github/workflows/build-macos.yml` and `.github/workflows/build-windows.yml`
+- Makefile targets: `build-r-macos`, `test-r-macos`, `build-r-windows`, `test-r-windows`
+
 ## Code style
 
 - Shell scripts use `bash` (manylinux) or `sh` (musllinux/Alpine)
--- a/README.md
+++ b/README.md
@@ -431,8 +431,15 @@ sudo ln -s ~/R/R-${R_VERSION} \
 
 Don't run this on a host that already has a real R install at that
 version-arch — it'll shadow the framework `Resources` for that version.
-RStudio and command-line use don't need this step. See
-[`macos/README.md`](macos/README.md) for the technical reasoning.
+RStudio and command-line use don't need this step.
+
+The architectures of Positron and the R install must match: Positron's R
+kernel loads `libR.dylib` in-process, and macOS dyld cannot `dlopen` a
+dylib of a different architecture. On Apple Silicon, install x86_64
+Positron alongside arm64 if you need to use the x86_64 portable R build,
+or stick with the arm64 build. RStudio does not have this constraint
+because it launches R as a subprocess. See [`macos/README.md`](macos/README.md)
+for the technical reasoning.
 
 #### Portable Windows (experimental)
 
--- a/README.md
+++ b/README.md
@@ -34,13 +34,18 @@ will be discontinued, but existing binaries will continue to be available.
 
 ### Portable builds (experimental)
 
-Portable R builds are also available that work across Linux distributions
-without distro-specific packages. These builds bundle most library dependencies
-and are relocatable to any install path.
+Portable R builds are also available that bundle most library dependencies
+and are relocatable to any install path. The Linux variants work across
+distributions without distro-specific packages; the macOS and Windows
+variants post-process the official CRAN binaries so they can be extracted
+anywhere instead of installed system-wide.
 
 - **manylinux** - any Linux distro with glibc >= 2.34 (RHEL 9+, Ubuntu 22.04+,
   Debian 12+, Amazon Linux 2023+, Arch Linux, etc.)
 - **musllinux** - Alpine Linux 3.20+ and other musl-based distros
+- **macOS** - arm64 (Apple Silicon) and x86_64 (Intel; also runs under
+  Rosetta 2), R 4.1.0+
+- **Windows** - x86_64, R 3.6.3+
 
 ## Supported R Versions
 
@@ -353,28 +358,91 @@ sudo apk add --no-cache \
 ```
 
 
-#### macOS
+#### Portable macOS (experimental)
+
+Portable, relocatable macOS R builds are available for arm64 (Apple Silicon)
+and x86_64 (Intel). These are post-processed CRAN binaries: the official
+`.pkg` installer is downloaded, extracted without installing, and patched so
+all hardcoded `/Library/Frameworks/R.framework/...` paths in Mach-O load
+commands and config files are rewritten to `@rpath`/`@loader_path`
+references. The result is a tarball that can be extracted to any directory
+and run from there — no admin rights, no Gatekeeper installer prompts, no
+side effects on the system R installation.
+
+These macOS builds are available for R 4.1.0 and later. R 4.0.x and R 3.x
+are not supported because CRAN does not host a macOS `.pkg` installer for
+those versions on either the main mirror or the CRAN archive. Use the
+official CRAN installer for those versions instead.
+
+Bundled libraries (Tcl/Tk, gfortran runtime, etc.) come from the CRAN `.pkg`,
+so behavior matches CRAN's official binary R for macOS. Each `.so` and
+`.dylib` is signed (ad-hoc on staging, Developer ID + notarized on
+production); when the production secrets are configured, downloaded tarballs
+pass Gatekeeper without quarantine. See [`macos/README.md`](macos/README.md)
+for the full technical breakdown.
+
+##### Install via tarball
 
 ```bash
-# arm64 (Apple Silicon)
-curl -LO https://cdn.posit.co/r/macos-arm64/R-4.4.1-macos-arm64.tar.gz
-tar xzf R-4.4.1-macos-arm64.tar.gz
-./R-4.4.1/bin/R --version
+R_VERSION=4.4.3
 
-# x86_64 (Intel)
-curl -LO https://cdn.posit.co/r/macos-x86_64/R-4.4.1-macos-x86_64.tar.gz
-tar xzf R-4.4.1-macos-x86_64.tar.gz
-./R-4.4.1/bin/R --version
+# arm64 (Apple Silicon)
+curl -O https://cdn.posit.co/r/macos-arm64/R-${R_VERSION}-macos-arm64.tar.gz
+mkdir -p ~/R
+tar xzf R-${R_VERSION}-macos-arm64.tar.gz -C ~/R
+~/R/R-${R_VERSION}/bin/R --version
+
+# x86_64 (Intel; also runs under Rosetta 2 on Apple Silicon)
+curl -O https://cdn.posit.co/r/macos-x86_64/R-${R_VERSION}-macos-x86_64.tar.gz
+mkdir -p ~/R
+tar xzf R-${R_VERSION}-macos-x86_64.tar.gz -C ~/R
+~/R/R-${R_VERSION}/bin/R --version
 ```
 
-#### Windows
+If you downloaded an unsigned/un-notarized tarball with `curl`, macOS
+attaches a quarantine attribute on extracted files. Strip it once with:
+
+```bash
+xattr -dr com.apple.quarantine ~/R/R-${R_VERSION}
+```
+
+Optional — for installing R packages that require compilation from source,
+install the Xcode Command Line Tools:
+
+```bash
+xcode-select --install
+```
+
+#### Portable Windows (experimental)
+
+Portable, relocatable Windows R builds are available for x86_64. The
+official CRAN `.exe` installer is downloaded and extracted with
+[`innoextract`](https://github.com/dscharrer/innoextract) (with a
+`/VERYSILENT /CURRENTUSER` silent-install fallback if `innoextract` is
+unavailable), so no admin rights, no registry changes, and no side effects
+on the system R installation. Windows R is already largely self-contained
+(all DLLs bundled, paths largely relative), so unlike macOS no Mach-O
+patching pipeline is needed — extraction + a portable site-library and
+default repo configuration is all that's required.
+
+These Windows builds are available for R 3.6.3 and later, with R 3.6.3
+included as a long-term compatibility anchor (matching the existing Linux
+builds). See [`windows/README.md`](windows/README.md) for the full technical
+breakdown.
+
+##### Install via zip
 
 ```powershell
-# Download and extract
-Invoke-WebRequest -Uri https://cdn.posit.co/r/windows/R-4.4.1-windows.zip -OutFile R-4.4.1-windows.zip
-Expand-Archive R-4.4.1-windows.zip -DestinationPath .
-.\R-4.4.1\bin\R.exe --version
+$RVersion = "4.4.3"
+
+Invoke-WebRequest -Uri "https://cdn.posit.co/r/windows/R-$RVersion-windows.zip" -OutFile "R-$RVersion-windows.zip"
+Expand-Archive "R-$RVersion-windows.zip" -DestinationPath C:\
+& "C:\R-$RVersion\bin\R.exe" --version
 ```
+
+Optional — for installing R packages that require compilation from source,
+install [Rtools](https://cran.r-project.org/bin/windows/Rtools/) (the
+version that matches your R minor version, e.g. Rtools 4.4 for R 4.4.x).
 
 ### Verify R installation
 
--- a/README.md
+++ b/README.md
@@ -413,6 +413,27 @@ install the Xcode Command Line Tools:
 xcode-select --install
 ```
 
+**Using the portable R inside Positron** — Positron's R discovery requires
+each R installation to live at the canonical
+`/Library/Frameworks/R.framework/Versions/<ver>-<arch>/Resources` path on
+disk (or be reachable via that path through a symlink). For the portable R
+to appear in Positron's interpreter picker, either install the tarball
+directly at that path, or symlink it once after extraction:
+
+```bash
+ARCH=arm64       # or x86_64
+RVER_MM=4.4      # major.minor of the R you extracted
+
+sudo mkdir -p /Library/Frameworks/R.framework/Versions/${RVER_MM}-${ARCH}
+sudo ln -s ~/R/R-${R_VERSION} \
+  /Library/Frameworks/R.framework/Versions/${RVER_MM}-${ARCH}/Resources
+```
+
+Don't run this on a host that already has a real R install at that
+version-arch — it'll shadow the framework `Resources` for that version.
+RStudio and command-line use don't need this step. See
+[`macos/README.md`](macos/README.md) for the technical reasoning.
+
 #### Portable Windows (experimental)
 
 Portable, relocatable Windows R builds are available for x86_64. The
--- a/README.md
+++ b/README.md
@@ -353,6 +353,29 @@ sudo apk add --no-cache \
 ```
 
 
+#### macOS
+
+```bash
+# arm64 (Apple Silicon)
+curl -LO https://cdn.posit.co/r/macos-arm64/R-4.4.1-macos-arm64.tar.gz
+tar xzf R-4.4.1-macos-arm64.tar.gz
+./R-4.4.1/bin/R --version
+
+# x86_64 (Intel)
+curl -LO https://cdn.posit.co/r/macos-x86_64/R-4.4.1-macos-x86_64.tar.gz
+tar xzf R-4.4.1-macos-x86_64.tar.gz
+./R-4.4.1/bin/R --version
+```
+
+#### Windows
+
+```powershell
+# Download and extract
+Invoke-WebRequest -Uri https://cdn.posit.co/r/windows/R-4.4.1-windows.zip -OutFile R-4.4.1-windows.zip
+Expand-Archive R-4.4.1-windows.zip -DestinationPath .
+.\R-4.4.1\bin\R.exe --version
+```
+
 ### Verify R installation
 
 Test that R was successfully installed by running:
--- a/macos/install-rprofile-hook.sh
+++ b/macos/install-rprofile-hook.sh
@@ -1,29 +1,56 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# install-rprofile-hook.sh -- Install portable-R startup hooks into the base
+# Rprofile so they apply in every R launch context.
+#
+# Why the base Rprofile and not etc/Rprofile.site?
+#   etc/Rprofile.site is skipped under `R --vanilla` and is also bypassed
+#   when an embedding host (RStudio's rsession, Positron's R kernel) loads
+#   libR.so directly without going through bin/R. The base Rprofile at
+#   library/base/R/Rprofile is sourced by R itself during startup, so it
+#   runs in every launch context.
+#
+# This file appends three local() blocks to the base Rprofile:
+#   1. Default CRAN repo -> p3m.dev (Posit Public Package Manager)
+#   2. TCL_LIBRARY / TK_LIBRARY env vars pointing at our bundled lib/tcl8.6
+#      and lib/tk8.6 directories (R < 4.3 needs these; R >= 4.3 ignores them
+#      because Tcl/Tk is bundled inside R.framework)
+#   3. A .portable environment that wraps install.packages() to rewrite
+#      hardcoded /Library/Frameworks/R.framework/... load commands in
+#      newly-installed CRAN binary packages. This is macOS-specific and
+#      gated on Sys.info()[["sysname"]] == "Darwin".
+#
+# The existing Rprofile content from CRAN is preserved (we append, not
+# overwrite). bin/fix-dylibs is also installed as a manual escape hatch
+# in case the install.packages wrapper is shadowed by an IDE.
+
 R_HOME="${1:?Usage: install-rprofile-hook.sh <r-home>}"
 R_HOME="$(cd "${R_HOME}" && pwd)"
 
-echo "=== Installing Rprofile.site hook in ${R_HOME} ==="
+BASE_RPROFILE="${R_HOME}/library/base/R/Rprofile"
+if [[ ! -f "${BASE_RPROFILE}" ]]; then
+  echo "ERROR: base Rprofile not found at ${BASE_RPROFILE}" >&2
+  exit 1
+fi
 
-cat > "${R_HOME}/etc/Rprofile.site" << 'RPROFILE'
-# Rprofile.site — rstudio/r-builds portable R distribution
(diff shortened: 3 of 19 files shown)


jmwoliver:
For notarizing, I think we need these five secrets:
| Secret | Content |
  | --- | --- |
  | `MACOS_DEVELOPER_CERTIFICATE` | base64 of your `.p12` (Developer ID Application cert + private key) |
  | `MACOS_DEVELOPER_CERTIFICATE_PASSWORD` | the `.p12` export password |
  | `MACOS_NOTARIZATION_USER_NAME` | your Apple ID email |
  | `MACOS_NOTARIZATION_USER_PASSWORD` | app-specific password from appleid.apple.com → Sign-In and Security |
  | `APPLE_TEAM_ID` | 10-char Team ID from developer.apple.com → Membership |

macos/install-rprofile-hook.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+R_HOME="${1:?Usage: install-rprofile-hook.sh <r-home>}"
+R_HOME="$(cd "${R_HOME}" && pwd)"
+
+echo "=== Installing Rprofile.site hook in ${R_HOME} ==="
+
+cat > "${R_HOME}/etc/Rprofile.site" << 'RPROFILE'
```
glin: I learned earlier that `R_HOME/etc/Rprofile.site` does not work if you launch R with `--vanilla`, as it gets ignored in that case. But the base Rprofile does work, e.g. see: https://github.com/rstudio/r-builds/blob/1ee9572faba1b4555e5950f32fa3033fdd6d56bd/builder/package.manylinux_2_34#L114-L118

macos/install-rprofile-hook.sh
```diff
@@ -0,0 +1,133 @@
...
+# Rprofile.site — rstudio/r-builds portable R distribution
+# This file is sourced at R startup. It provides:
+# 1. Automatic patching of CRAN binary packages for macOS portability
+# 2. Posit Package Manager as default repository
+
+# ── Default repository: Posit Package Manager ──────────────────────
+local({
+  r <- getOption("repos")
+  r["CRAN"] <- "https://packagemanager.posit.co/cran/latest"
```
glin: This is a difference vs. the existing R builds, which don't change the default CRAN mirror. I don't see a problem with it since r-builds is billed as opinionated, and these are experimental, but just noting that there would be an inconsistency.

But.. there's a good chance we'll eventually switch over the Linux R builds to use P3M as the default CRAN repo, especially once we have native Linux binary support.

macos/install-rprofile-hook.sh
```diff
@@ -0,0 +1,133 @@
...
+
+  # Wrapper around utils::install.packages that patches .so files after install
+  .portable$install.packages <- function(pkgs, ...) {
+    utils <- asNamespace("utils")
+    result <- utils$install.packages(pkgs, ...)
+    mc <- match.call()
+    lib_dir <- if ("lib" %in% names(mc)) eval(mc$lib) else .libPaths()[1L]
+    try(.portable$fix_pkgs(pkgs, lib = lib_dir), silent = TRUE)
+    invisible(result)
+  }
```
glin, line 127: I'm curious if this works in the RStudio IDE since the IDE installs its own hook that overwrites `install.packages()`

macos/install-rprofile-hook.sh
glin: Side note: p3m.dev because I think we've come to prefer that generally
```suggestion
  r["CRAN"] <- "https://p3m.dev/cran/latest"
```

README.md
```diff
@@ -353,6 +353,29 @@ sudo apk add --no-cache \
...
+./R-4.4.1/bin/R --version
+
+# x86_64 (Intel)
+curl -LO https://cdn.posit.co/r/macos-x86_64/R-4.4.1-macos-x86_64.tar.gz
+tar xzf R-4.4.1-macos-x86_64.tar.gz
+./R-4.4.1/bin/R --version
+```
+
+#### Windows
+
```
glin: I would add more detail here, like which R versions are available, a blurb about being portable, experimental status, etc

CLAUDE.md
```diff
@@ -39,6 +39,13 @@ See [builder/patches/README.md](builder/patches/README.md).
 - **OpenBLAS naming inconsistency**: OpenBLAS .so names differ across distros (`libopenblaso.so` on Rocky 9 vs `libopenblas.so` on Ubuntu/Alpine). This is why portable builds use a post-build BLAS swap (keeping Makeconf's `-lRblas`) rather than `--with-blas=<lib>` at configure time.
 - **PCRE2 hidden for R 3.x**: `build.sh` temporarily hides PCRE2 pkg-config during configure for R < 4.0 to prevent unwanted linkage.
 
+## macOS and Windows builds
+
+- macOS build scripts live in `macos/`, Windows scripts in `windows/`
+- These use CRAN binaries + post-processing (not compiled from source like Linux builds)
+- CI workflows are `.github/workflows/build-macos.yml` and `.github/workflows/build-windows.yml`
+- Makefile targets: `build-r-macos`, `test-r-macos`, `build-r-windows`, `test-r-windows`
```
glin, line 48: I would also add a more detailed internal README similar to the PR description or https://github.com/rstudio/r-builds/blob/main/builder/portable-r/README.md, to help us understand how this works. Like if R 4.7 releases and breaks something here, a technical overview would help to know how to fix that.

glin (commented):
The approach sounds good to me, and I'm up for testing it out. Do you want to build some test packages first using the GHA workflow?

In particular, I think we should test this through the RStudio IDE and Positron since I ran into issues earlier with RStudio/Positron running R in special ways that broke on the initial manylinux attempts.

jmwoliver:
@glin Thanks for the detailed review! I tested end to end with R 4.4.3 and 4.6.0 on Mac ARM64 and x86_64. I haven't tested Windows yet. I've attached steps below that I used if you want to test as well. Here's a successful run of Mac and Windows arm and x86 builds: https://github.com/rstudio/r-builds/actions/runs/25069345617

Do you know why `opensuse-156` fails on this PR? https://github.com/rstudio/r-builds/actions/runs/25069329782/job/73445078018?pr=284


---

# IDE testing playbook — portable R for macOS and Windows

Manual testing steps for the portable macOS and Windows R builds from PR #284, focused on RStudio and Positron compatibility. Run after a GitHub Actions build completes to validate the resulting tarballs/zips against real IDE sessions.

The key thing this playbook is designed to surface: whether the `.portable` Rprofile hook (which patches CRAN binary `.tgz` files to fix hardcoded `/Library/Frameworks/R.framework` references) survives RStudio's and Positron's own `install.packages()` overrides.

---

## 1. Download the artifacts

After the GHA build run completes, fetch its artifacts to your laptop:

```bash
RUN_ID=<paste-the-run-id>
mkdir -p /tmp/rbuilds-test
gh run download "$RUN_ID" --repo rstudio/r-builds \
  --dir /tmp/rbuilds-test/

# Each matrix entry becomes its own subdir:
#   /tmp/rbuilds-test/R-4.4.3-macos-arm64/R-4.4.3-macos-arm64.tar.gz
#   /tmp/rbuilds-test/R-4.4.3-windows/R-4.4.3-windows.zip
ls -la /tmp/rbuilds-test/
```

Pick one R version to focus on first. **R 4.4.3** is a good default — current LTS-ish, exercises the R ≥ 4.3 Tcl/Tk-bundled-in-framework path.

---

## 2. macOS setup (arm64 first; repeat for x86_64 if/when arm64 passes)

```bash
ARCH=arm64    # or x86_64
RVER=4.4.3
ART=/tmp/rbuilds-test/R-${RVER}-macos-${ARCH}

mkdir -p /tmp/portable-R
tar xzf "$ART/R-${RVER}-macos-${ARCH}.tar.gz" -C /tmp/portable-R/

# CRAN ships a couple of dylibs (libtcl8.6.dylib, libtk8.6.dylib) as
# mode 0444; xattr -d requires write access to clear an attribute. Make
# everything writable for the owner before xattr.
chmod -R u+w /tmp/portable-R/R-${RVER}

# Strip Gatekeeper quarantine. Current builds are ad-hoc signed only —
# no Apple notarization yet, so anything downloaded via curl/gh is
# quarantined.
xattr -dr com.apple.quarantine /tmp/portable-R/R-${RVER}
```

---

## 3. CLI sanity check — do this BEFORE any IDE

This is the cheapest way to confirm the base-Rprofile hooks landed correctly outside any IDE complications. If it passes here but fails in RStudio, the IDE is the culprit. If it fails here, something is wrong with the build itself.

```bash
/tmp/portable-R/R-${RVER}/bin/R --vanilla --no-echo -e '
cat("R.home              =", R.home(), "\n")
cat("repo[CRAN]          =", getOption("repos")["CRAN"], "\n")
cat(".portable in search =", ".portable" %in% search(), "\n")
cat("find:               ", find("install.packages"), "\n")
cat("has fix_pkgs:       ", grepl("fix_pkgs", paste(deparse(body(install.packages)), collapse=" ")), "\n")
'
```

Expected output (for macOS):
```
R.home              = /private/tmp/portable-R/R-4.4.3
repo[CRAN]          = https://p3m.dev/cran/latest
.portable in search = TRUE
find:                .portable package:utils
has fix_pkgs:        TRUE
```

**Why not `environmentName(environment(install.packages))`?** That returns `""` for our wrapper, because the wrapper is a closure created inside a `local({...})` block in the base Rprofile and `local()` envs are anonymous. `find()` instead reports search-path positions by name, and the body inspection confirms the function at that position is ours (it contains the call to our `fix_pkgs` patcher) and not the stock `utils::install.packages`.

If any line is wrong, **stop and report back** before going to the IDE. There is no point IDE-testing a broken base.

---

## 4. RStudio — macOS

```bash
RSTUDIO_WHICH_R=/tmp/portable-R/R-${RVER}/bin/R open -a RStudio
```

Persistent alternative: **Tools → Global Options → General → R version → "Change…" → "Choose a specific version of R"** → browse to `/tmp/portable-R/R-${RVER}/bin/R` → restart RStudio.

In RStudio's R console, paste **the diagnostic script in section 6**.

---

## 5. RStudio — Windows

```powershell
$RVer = "4.4.3"
Expand-Archive `
  "$HOME\Downloads\rbuilds-test\R-$RVer-windows\R-$RVer-windows.zip" `
  -DestinationPath C:\portable-R -Force

# Tell RStudio which R to use
$env:RSTUDIO_WHICH_R = "C:\portable-R\R-$RVer\bin\x64\Rgui.exe"
& "C:\Program Files\RStudio\rstudio.exe"
```

Persistent: same Tools → Global Options path, browse to `C:\portable-R\R-${RVer}\bin\x64\Rgui.exe`.

In RStudio's R console, paste the diagnostic script in section 6.

---

## 6. The diagnostic script

Paste the **whole block** into the IDE's R console. Don't edit it, even if some output looks wrong — the script is designed so the failures are diagnostic.

```r
# ── 1. Where is R running from? ─────────────────────────────────
cat("R.home() =", R.home(), "\n")
cat("R version:", R.version.string, "\n\n")

# ── 2. Did the base-Rprofile hooks survive RStudio's startup? ───
cat("repo[CRAN] =", getOption("repos")["CRAN"], "\n")
cat("\nsearch() path:\n"); print(search())

# ── 3. Whose install.packages() are we calling? ─────────────────
# environmentName() returns "" for anonymous local() envs, which is
# what our .portable hook lives inside, so we use find() + body
# inspection instead.
cat("\nfind('install.packages') =", find("install.packages"), "\n")
cat("body has fix_pkgs (our marker) =",
    grepl("fix_pkgs", paste(deparse(body(install.packages)), collapse=" ")),
    "\n")
# Interpretation of find() output (first match wins):
#   ".portable"      our hook is winning (best case on macOS)
#   "tools:rstudio"  RStudio's wrapper is winning, .portable is shadowed
#   "package:utils"  neither override is in front (problem on macOS)
#   <other>          something unexpected

# ── 4. Install a package and check whether dylibs got patched ───
tmp <- tempfile("rstudio-test-lib-"); dir.create(tmp)
cat("\ninstalling jsonlite into:", tmp, "\n")
install.packages("jsonlite", lib = tmp, type = "binary")
cat("\nattempting to load it:\n")
ok <- tryCatch({ library(jsonlite, lib.loc = tmp); TRUE },
               error = function(e) { cat("LOAD ERROR:", conditionMessage(e), "\n"); FALSE })
cat("library(jsonlite) success:", ok, "\n")

# ── 5. (macOS only) Did the framework path get rewritten? ───────
if (Sys.info()[["sysname"]] == "Darwin") {
  so <- list.files(file.path(tmp, "jsonlite", "libs"),
                   pattern = "\\.so$", full.names = TRUE)
  if (length(so)) {
    cat("\nchecking dylib refs in:", so, "\n")
    refs <- system2("otool", c("-L", shQuote(so)), stdout = TRUE)
    fw   <- grep("/Library/Frameworks/R.framework", refs, value = TRUE)
    if (length(fw)) {
      cat("UNPATCHED — these refs need fixing:\n")
      cat(paste0("  ", fw), sep = "\n")
      cat("\nrun this to fix manually:\n")
      cat("  ", file.path(R.home(), "bin/fix-dylibs"), "\n")
    } else {
      cat("dylib clean — no /Library/Frameworks/R.framework refs\n")
    }
  }
}

# ── 6. Plotting smoke test ──────────────────────────────────────
cat("\n--- plotting test ---\n")
plot(1:10)                                      # cairo / Quartz device
png_path <- tempfile(fileext = ".png")
png(png_path); plot(rnorm(100)); dev.off()
cat("png written to:", png_path,
    "size:", file.info(png_path)$size, "bytes\n")
```

---

## 7. Positron — macOS

Positron's macOS R discovery requires that the parsed `R_HOME_DIR=` value (read from `bin/R` as text) point at a path that exists on disk with `library/utils/DESCRIPTION` inside. Our portable R's static line is the orthogonal `Versions/<ver>-<arch>/Resources` path, so end users either install at that canonical path *or* symlink to the portable extraction. The same applies for testing — symlink once and Positron's standard discovery walks the framework `Versions/` directory and picks up the install:

```bash
sudo mkdir -p /Library/Frameworks/R.framework/Versions/4.4-arm64
sudo ln -s /tmp/portable-R/R-${RVER} \
  /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources
```

⚠️ **Don't run this on a host that already has a real R 4.4 install** — it will shadow the framework's `Resources` and break the system R. For test machines without R 4.4 installed, it's safe.

No `customBinaries` or `customRootFolders` setting is needed once the symlink is in place — discovery picks up the install via the standard framework path.

Then open Positron normally:

1. Launch Positron (`open -a Positron`). The portable R should now appear under the standard discovery path — no `customBinaries`/`customRootFolders` setting needed.
2. **⌘⇧P** → `R: Select Interpreter` → pick `R 4.4.3` from the list (it'll show as living at `/Library/Frameworks/R.framework/Versions/4.4-arm64/Resources` per the symlink).
3. Verify the session header reads the expected version and platform (`R version 4.4.3` / `aarch64-apple-darwin20`) before running the diagnostic.
4. Paste the diagnostic from section 6.

**Debugging discovery:** if the portable R still doesn't appear in the picker, open the Output panel (**⇧⌘U**) and select the **"R Language Pack"** channel from the dropdown. Discovery decisions are logged there with concrete rejection reasons.

**Cleanup after testing:**
```bash
sudo rm /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources
sudo rmdir /Library/Frameworks/R.framework/Versions/4.4-arm64
```

In Positron's R console, paste the diagnostic script from section 6.

---

## 8. Positron — Windows

Same approach as macOS — use `customBinaries` to point Positron at the portable R directly:

1. Launch Positron normally.
2. **Ctrl+Shift+P** → `Preferences: Open User Settings (JSON)` and add:

   ```json
   "positron.r.customBinaries": ["C:\\portable-R\\R-4.4.3\\bin\\x64\\R.exe"]
   ```

   (Note the doubled backslashes — required JSON escaping on Windows paths. Adjust `R-4.4.3` for your version.)

3. **Ctrl+Shift+P** → `Developer: Reload Window`.
4. **Ctrl+Shift+P** → `R: Select Interpreter` → pick the portable R from the list.
5. Verify the session header reports the expected version before running the diagnostic.

If discovery doesn't pick it up, open **View → Output** and select the **"R Language Pack"** channel from the dropdown to see why it was rejected.

Then paste the diagnostic from section 6.

---

## 9. What to send back

Run the diagnostic in **at least these two contexts** to start:

1. **macOS arm64 + RStudio**
2. **macOS arm64 + Positron**

For each, paste the entire console output (including any errors). The four data points to look for:

| Question | Where in the output |
|---|---|
| Is `.portable` in the search path? | Section 2's `search()` print |
| Whose `install.packages()` ran? | Section 3 `find('install.packages')` first entry + `body has fix_pkgs` line |
| Did `library(jsonlite)` load cleanly? | Section 4 `library(jsonlite) success:` line |
| Did the dylib get patched? | Section 5 `dylib clean` vs `UNPATCHED` |

Those four are enough to decide whether the `.portable` hook needs more work, whether `bin/fix-dylibs` is the right escape hatch to document, or whether the base-Rprofile move alone solved it end-to-end.

Windows results are interesting too but lower priority — Windows R doesn't have the dylib patching problem, so the only Windows checks are "does the IDE see and run the portable R" plus "does PPM resolve as the default repo". If Windows passes the CLI sanity check in section 3, it is almost certainly fine in the IDEs.

---

## Interpretation cheat sheet

| Scenario | Interpretation | Action |
|---|---|---|
| All four checks PASS in both RStudio and Positron | The base-Rprofile move alone solved the IDE problem. | Update `macos/README.md` to remove the "we'll revisit" caveat about IDE shadowing. |
| `.portable` in `search()` but `install.packages` resolves to `tools:rstudio` (RStudio shadows it) | Expected outcome if RStudio's tools attach above `.portable`. | Iterate: re-attach `.portable` after RStudio's startup completes, or wrap `dyn.load` instead of `install.packages` so the fix-up runs at package-load time. |
| `library(jsonlite)` errors with "image not found" but `bin/fix-dylibs` cleans it up | The wrapper was bypassed but the manual recovery works. | Keep the current design, document `bin/fix-dylibs` as the recovery path more prominently. |
| `library(jsonlite)` errors and `bin/fix-dylibs` does **not** clean it up | Something deeper than IDE wrapping — possibly a relocation or codesign bug. | Send back `otool -L` output of the failing `.so`. |
| Windows passes CLI but RStudio can't find R | RStudio's bin/R parser is rejecting our extracted layout. | Send back `Get-Content C:\portable-R\R-$RVer\bin\R.exe -TotalCount 20` so we can see what RStudio is reading. |

macos/install-rprofile-hook.sh
jmwoliver: Yeah I thought I did it for a reason like a specific R version / platform / architecture only existed on P3M and not provided by CRAN, but I don't remember. I agree I still think it make sense. I'd definitely make sense for manylinux R build to get the manylinux binaries we provide right?

jmwoliver: Good point, updated.

macos/install-rprofile-hook.sh
jmwoliver: I had to do a few updates to get things working, but RStudio and Positron work even with the hooks that they add. I provided steps below for what I tested.

macos/install-rprofile-hook.sh
jmwoliver: Good catch, I was able to get this matching manylinux and add a test to verify `--vanilla` works. I also added manual testing steps to show what all I tested in the CLI, Positron, and RStudio.

README.md
jmwoliver: Updated with more detail, let me know if that looks better.

CLAUDE.md
jmwoliver: I added more detailed macos and windows READMEs, let me know if those look better.

glin:
@jmwoliver SUSE mirrors have always been incredibly flaky, so I usually just rerun those when they fail.

I'll give it a test run on my Windows system when I get a chance.

glin:
I just tested this on Windows and it works great. I tried 3.6 to devel, in Positron and RStudio, Powershell and Git Bash.

For Positron testing, it's easier to point `Positron › R › Interpreters: Override` to a directory containing one or more R installations, since Custom Binaries expects specific paths.

glin (approved):
Is there anything left to do for macOS? The notarization part?

jmwoliver:
Oh cool I didn't know about that Positron setting. For more detail since I wasn't familiar, I was able to add this to the user settings:
```
"positron.r.interpreters.override": [
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-4.2.3/bin/R",
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-4.3.3/bin/R",
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-4.4.3/bin/R",
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-4.5.3/bin/R",
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-4.6.0/bin/R",
    "/Users/jacobwoliver/r-builds-artifacts/macos-arm64/extracted/R-devel/bin/R"
  ]
```

And that got all the custom R builds to be detected:
Screenshot 2026-04-30 at 9 29 54 AM

jmwoliver:
@glin Yes just the notarization is left. Do you think this PR is okay to merge with the caveat that these builds are experimental and you need to un-quarantine these Mac R builds manually until we get notarization in place?

I can work with CloudOps to see if we can get this working. Here's the existing Slack thread for context: https://positpbc.slack.com/archives/CKJTZMWHM/p1776969569044419?thread_ts=1776965096.651149&cid=CKJTZMWHM

macos/make-relocatable.sh
```diff
@@ -5,17 +5,47 @@ R_HOME="${1:?Usage: make-relocatable.sh <r-home> [r-version]}"
...
-# in that case (and when R_VERSION is unset entirely).
-if [[ -n "${R_VERSION}" ]] && [[ "${R_VERSION}" =~ ^[0-9]+\.[0-9]+ ]]; then
-  R_MAJOR="${R_VERSION%%.*}"
-  R_MINOR="${R_VERSION#*.}"; R_MINOR="${R_MINOR%%.*}"
-else
-  R_VER_STRING=$("${R_HOME}/bin/exec/R" --version 2>&1 | head -1 | grep -oE '[0-9]+\.[0-9]+' | head -1 || echo "")
-  R_MAJOR="${R_VER_STRING%%.*}"
-  R_MINOR="${R_VER_STRING#*.}"; R_MINOR="${R_MINOR%%.*}"
-fi
+# Determine the version slot used in the orthogonal R_HOME_DIR= rewrite
```
jmwoliver, line 8: I think this hack is only needed because Positron requires the path to actually exist to fetch the DESCRIPTION or something. So we have to set the path to the standard Library path and then symlink. We could probably remove this if we make a change in Positron so that it can use the relative pathing.

This is the current path in Positron:
1. Read script lines from binpath (existing).
2. Verify "Shell wrapper for R executable" header (existing).
3. Parse R_HOME_DIR (existing).
4. If parsed value validates (DESCRIPTION exists at <homepath>/library/utils/DESCRIPTION) → return it.

Step 4 is where it breaks currently. We would want to add the following (which I think Windows already does):

  6. Else fall back to dirname(dirname(realpath(binpath))). If that validates → return it.
  7. Else return undefined (today's behavior on failure).

jmwoliver:
There was a small issue parsing the devel/patched/next version for Mac, that is fixed now.

glin:
@jmwoliver I think fine to merge with notarization added later.

Also for the interpreters setting, it discovers R installations recursively, so you can just set it to the `/macos-arm64/extracted/` path to discover all R versions at once.

PR #292, opened 2026-04-30
Unify macOS CDN path with Linux convention

jmwoliver:
## Summary

Move both macOS arches under a single `/r/macos/` prefix and adopt the same arch-suffix filename convention Linux already uses. macOS becomes structurally identical to Linux on the CDN.

**Before** (never published — landing change before the first publish):

```
/r/macos-arm64/R-{ver}-macos-arm64.tar.gz
/r/macos-x86_64/R-{ver}-macos-x86_64.tar.gz
```

**After**:

```
/r/macos/R-{ver}-macos.tar.gz          # x86_64
/r/macos/R-{ver}-macos-arm64.tar.gz    # arm64
```

The shape across all three OS families is now `/r/<platform>/R-<ver>-<platform>[-arm64].tar.gz`.

## Why this shape

The original PR #284 introduced macOS with a third pattern (one prefix per arch) that didn't match either Linux (one prefix per platform, arm64 suffix in filename) or any common precedent. While we're talking about whether to add arch to the Windows path for future-proofing, it became clearer that macOS was the thing actually out of step with the rest of the repo.

A few specific decisions that got made along the way:

- **One prefix instead of two.** Centralising under `/r/macos/` makes discovery (and any future mirror/listing tooling) cleaner — one prefix to enumerate, one place to look. This also aligns with the modern industry convention used by Go, Node.js, Rust, uv, and most GitHub Releases: all platform/arch variants in one directory, the filename does the disambiguating. Older ecosystems like apt/CRAN-macOS use arch-in-directory because their package metadata format requires it; we don't have that constraint.
- **`macos`, not `macosx`.** CRAN uses `bin/macosx/...`, but that's a legacy name from the OS X era (Apple dropped the X in 2012). Within this repo every other platform identifier is a current name (`ubuntu-2204`, `rhel-9`, `windows`), so `macosx` would be the odd one out. This isn't a CRAN mirror — within-repo consistency wins over CRAN parity.
- **Keep the platform in the filename.** `R-4.4.3-macos.tar.gz` rather than `R-4.4.3.tar.gz` — once a tarball is sitting in `~/Downloads/`, the filename alone should still tell you what it is. Linux already does this, and it disambiguates from the source tarball, the manylinux portable, etc.
- **x86_64 unsuffixed, arm64 suffixed.** Matching Linux. Both arches are first-class on macOS in 2026, but adopting the existing repo convention beats inventing a new one for this platform.

## Why this is the right time

Nothing has been published from PR #284 yet — the production CDN currently 404s on `/r/macos-arm64/...` and `/r/macos-x86_64/...`. Renaming now is free; renaming after the first publish would be a documented public-URL break.

## Files changed

- `macos/build.sh` — tarball naming uses arch-suffix convention (x86_64 unsuffixed, arm64 keeps `-arm64`)
- `Makefile` — `test-r-macos` extracts the right filename per arch
- `.github/workflows/build-macos.yml` — build-step tarball var + S3 key both updated
- `macos/README.md` — package-name and CDN-URL block
- `README.md` — install snippets

## Intentionally unchanged

- **GHA artifact folder names** stay `R-{ver}-macos-{arch}` so x86_64 and arm64 don't collide in the GHA UI; the file inside the artifact follows the new CDN naming. (Two artifacts both named `R-4.4.3-macos.tar.gz` would be confusing on the run page.)
- **User-input platform selectors** in `build.yml` (`macos-arm64`, `macos-x86_64`) drive matrix filtering for `workflow_dispatch`. They're not S3 paths — separate concern, no reason to churn them.

## Windows

Windows stays as-is for now (`/r/windows/R-{ver}-windows.zip`). When Windows arm64 R is ready, the same Linux-style suffix can be added without a path migration: `/r/windows/R-{ver}-windows-arm64.zip`.

## Test plan

- [ ] CI passes on this branch
- [ ] Manual `make build-r-macos R_VERSION=4.4.3 ARCH=arm64` produces `output/R-4.4.3-macos-arm64.tar.gz`
- [ ] Manual `make build-r-macos R_VERSION=4.4.3 ARCH=x86_64` produces `output/R-4.4.3-macos.tar.gz`
- [ ] `make test-r-macos R_VERSION=4.4.3 ARCH=arm64` extracts `R-4.4.3-macos-arm64.tar.gz`
- [ ] `make test-r-macos R_VERSION=4.4.3 ARCH=x86_64` extracts `R-4.4.3-macos.tar.gz`
- [ ] Staging publish lands artifacts at the new `/r/macos/...` prefix in the staging bucket

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ anywhere instead of installed system-wide.
 - **manylinux** - any Linux distro with glibc >= 2.34 (RHEL 9+, Ubuntu 22.04+,
   Debian 12+, Amazon Linux 2023+, Arch Linux, etc.)
 - **musllinux** - Alpine Linux 3.20+ and other musl-based distros
-- **macOS** - arm64 (Apple Silicon) and x86_64 (Intel; also runs under
-  Rosetta 2), R 4.1.0+
+- **macOS** - arm64 (Apple Silicon) and x86_64 (Intel), R 4.1.0+
 - **Windows** - x86_64, R 3.6.3+
 
 ## Supported R Versions
@@ -392,7 +391,7 @@ mkdir -p ~/R
 tar xzf R-${R_VERSION}-macos-arm64.tar.gz -C ~/R
 ~/R/R-${R_VERSION}/bin/R --version
 
-# x86_64 (Intel; also runs under Rosetta 2 on Apple Silicon)
+# x86_64 (Intel)
 curl -O https://cdn.posit.co/r/macos/R-${R_VERSION}-macos.tar.gz
 mkdir -p ~/R
 tar xzf R-${R_VERSION}-macos.tar.gz -C ~/R
--- a/README.md
+++ b/README.md
@@ -387,15 +387,15 @@ for the full technical breakdown.
 R_VERSION=4.4.3
 
 # arm64 (Apple Silicon)
-curl -O https://cdn.posit.co/r/macos-arm64/R-${R_VERSION}-macos-arm64.tar.gz
+curl -O https://cdn.posit.co/r/macos/R-${R_VERSION}-macos-arm64.tar.gz
 mkdir -p ~/R
 tar xzf R-${R_VERSION}-macos-arm64.tar.gz -C ~/R
 ~/R/R-${R_VERSION}/bin/R --version
 
 # x86_64 (Intel; also runs under Rosetta 2 on Apple Silicon)
-curl -O https://cdn.posit.co/r/macos-x86_64/R-${R_VERSION}-macos-x86_64.tar.gz
+curl -O https://cdn.posit.co/r/macos/R-${R_VERSION}-macos.tar.gz
 mkdir -p ~/R
-tar xzf R-${R_VERSION}-macos-x86_64.tar.gz -C ~/R
+tar xzf R-${R_VERSION}-macos.tar.gz -C ~/R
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -387,15 +387,15 @@ for the full technical breakdown.
 R_VERSION=4.4.3
 
 # arm64 (Apple Silicon)
-curl -O https://cdn.posit.co/r/macos-arm64/R-${R_VERSION}-macos-arm64.tar.gz
+curl -O https://cdn.posit.co/r/macos/R-${R_VERSION}-macos-arm64.tar.gz
 mkdir -p ~/R
 tar xzf R-${R_VERSION}-macos-arm64.tar.gz -C ~/R
 ~/R/R-${R_VERSION}/bin/R --version
 
 # x86_64 (Intel; also runs under Rosetta 2 on Apple Silicon)
```
glin: Side note, this part might be redundant? Or at least, I can't think of any reason to do this when the arm64 build exists
```suggestion
# x86_64 (Intel)
```

glin (approved):

README.md
jmwoliver: That's fair. It was added because we build x86 on an ARM runner through Rosetta 2 and that just propagated to all the references. I'll remove it here and anywhere else it doesn't make sense to mention it.

PR #293, opened 2026-05-01
Publish r-builds CDN manifest with sha256 sidecars

jmwoliver:
## Summary

Adds a single machine-readable manifest at `cdn.posit.co/r/manifest.json` that enumerates every R build available on the CDN — portable manylinux/musllinux/macOS/Windows tarballs alongside every native Linux distro deb/rpm/apk — so third-party tooling can discover and verifiably download R builds without per-platform probing or filename guesswork.

Every artifact also gets a `<artifact>.sha256` sidecar published next to it in S3, in `sha256sum -c`-compatible format. Consumers can verify a download with one shell line:

```sh
curl -O https://cdn.posit.co/r/manylinux_2_34/R-4.5.3-manylinux_2_34.tar.gz
curl -O https://cdn.posit.co/r/manylinux_2_34/R-4.5.3-manylinux_2_34.tar.gz.sha256
sha256sum -c R-4.5.3-manylinux_2_34.tar.gz.sha256
```

`versions.json` is unchanged in content and shape — the manifest is purely additive. The new code regenerates `versions.json` with read-modify-write semantics so a consolidator run before the initial sidecar backfill is complete cannot shrink the legacy index.

## Schema

```json
{
  "schema_version": 1,
  "generated_at": "2026-05-01T12:00:00Z",
  "builds": [
    {
      "r_version": "4.5.3",
      "platform": "manylinux_2_34",
      "arch": "amd64",
      "url": "https://cdn.posit.co/r/manylinux_2_34/R-4.5.3-manylinux_2_34.tar.gz",
      "sha256": "abc123...",
      "size": 123456789
    }
  ]
}
```

`(r_version, platform, arch)` is the natural primary key. Portability is identified by the platform string (`manylinux_2_34`, `musllinux_1_2`, `macos`, `windows`) — no separate boolean field. Sort order: named channels (`devel`, `next`, `patched`) at top in fixed order, numeric versions newest-first, then platform ascending, then arch ascending. RPM/APK arch is normalized to the dpkg convention (`x86_64`→`amd64`, `aarch64`→`arm64`) so all Linux entries share one identifier per hardware target.

## What changed

- **`build_manifest.py`** — parses S3 keys, fetches sidecars, assembles the manifest, regenerates `versions.json`, uploads both. CLI: `python3 build_manifest.py write --s3-bucket BUCKET --cdn-base URL [--dry-run]`.
- **`test_build_manifest.py`** — 27 unit tests covering parsing across all artifact shapes (deb/rpm/apk/tarball/zip including hyphenated platform names like `rhel-9`, `ubuntu-2204`), manifest assembly, sort order, and URL construction.
- **`scripts/backfill_sha256.py`** — one-shot maintenance tool. Streams every artifact, computes sha256, uploads `<artifact>.sha256` if missing. Idempotent and parallel (ThreadPoolExecutor).
- **`.github/workflows/update-manifest.yml`** *(new)* — reusable workflow with `workflow_call` and `workflow_dispatch` triggers. Runs once at the end of every publish run.
- **`build.yml`, `build-macos.yml`, `build-windows.yml`** — each gains a sha256 sidecar step before the publish step. The three scattered `update-versions-json` jobs are removed and replaced with one end-of-run call to `update-manifest.yml` from `build.yml`. The `update-manifest` job uses `if: !cancelled() && inputs.publish != ''` so partial-success runs still produce an accurate manifest while user-cancelled runs skip.
- **Makefile** — splits `unit-test` into `unit-test-builder` (existing) and `unit-test-manifest` (new).
- **README.md** — documents the manifest URL and sha256 sidecars.

## Test plan

- [ ] Approve and merge.
- [ ] Run `python3 scripts/backfill_sha256.py --s3-bucket <staging-bucket>` against staging until completion. Verify with `aws s3 ls s3://<staging-bucket>/r/ --recursive | grep -c '\.sha256$'` matches the artifact count.
- [ ] Trigger `update-manifest.yml` against staging via `gh workflow run update-manifest.yml -f publish=staging`.
- [ ] Fetch `https://cdn-staging…/r/manifest.json` and inspect the schema, sort order, and a sampling of URLs.
- [ ] Verify `versions.json` on staging is unchanged in shape and unioned with previous content.
- [ ] Repeat backfill against production.
- [ ] Trigger `update-manifest.yml` against production.
- [ ] Verify `https://cdn.posit.co/r/manifest.json` resolves and parses, and `versions.json` is unaffected for existing consumers (`install.sh` quick installer, etc.).

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -49,7 +49,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.setup-matrix.outputs.matrix }}
-      r_versions: ${{ steps.setup-matrix.outputs.r_versions }}
     steps:
       - uses: actions/checkout@v4
 
@@ -84,7 +83,6 @@ jobs:
           #   2. arm64 is additionally unsupported on R < 4.1 (no arm64 .pkg),
           #      but rule 1 already covers this — kept as a defensive comment.
           includes="[]"
-          built_versions=""
           for version in $(echo "$r_versions" | tr ',' ' '); do
             case "${version}" in
               devel|patched|next)
@@ -98,7 +96,6 @@ jobs:
                 fi
                 ;;
             esac
-            built_versions="${built_versions:+${built_versions},}${version}"
             for arch in $archs; do
               # Both arches run on macos-14 (arm64). x86_64 R runs under
               # Rosetta 2, which is pre-installed on GitHub's macos-14 images.
@@ -113,10 +110,6 @@ jobs:
           matrix=$(echo "$includes" | jq -c '{"include": .}')
           echo "matrix=$matrix" >> $GITHUB_OUTPUT
           echo "Using matrix: $matrix"
-
-          # Only publish versions we actually built; filtering happens above.
-          r_versions_json=$(echo "$built_versions" | jq -Rc 'split(",") | map(select(length > 0))')
-          echo "r_versions=$r_versions_json" >> $GITHUB_OUTPUT
 
   build:
     needs: setup-matrix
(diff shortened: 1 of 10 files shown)


jmwoliver:
@glin Curious about your thoughts for this. Not really necessary, but would enable the ability to enumerate all builds for all R versions / platforms / architectures with one `cdn.posit.co/r/manifest.json` call.

glin:
I like it, although it may be partially unnecessary or overlapping with the packages being published to Cloudsmith (https://github.com/rstudio/r-builds/pull/270), which I believe would already handle integrity verification + additional GPG signature verification automatically. That's the plan going forward for the distro packages.

I'm not sure if we can or should publish the tarballs to Cloudsmith. If Cloudsmith supports raw tarball hosting, then it's probably a good idea long term for all the automatic benefits. But if not, then this would be great for the tarballs that get ad-hoc curl installed.

One more question that I'm unsure of is that https://github.com/rstudio/r-builds/pull/270 is going to add per-package versioning rather than just having a single "latest" package, so I'm not sure if that needs to be factored in here or not.

PR #304, opened 2026-05-14
Skip R-next Windows builds

nodivbyzero:
The daily devel workflow passes "devel,next" to all platforms, but CRAN's Windows binaries only include R-devel-win.exe, not R-next-win.exe. This caused Windows builds to fail with 404 errors when attempting to download the non-existent installer.

Filter out "next" from the Windows build matrix, matching how macOS already skips unsupported versions. Linux and macOS continue building both `devel` and `next` from source/nightly .pkg files.

Related to: https://github.com/rstudio/r-builds/actions/runs/25843407005/job/75933276364

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -56,8 +56,15 @@ jobs:
           echo "Resolved R versions: $r_versions"
 
           # Build matrix JSON (Windows is x86_64 only — no arch filtering needed)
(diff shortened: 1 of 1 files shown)


jmwoliver:
I think the R-next builds should be using the patched versions: https://github.com/rstudio/r-builds/pull/305. We just didn't have it configured properly.

nodivbyzero:
💯 Makes sense

PR #307, opened 2026-05-27
Remove downloaded installer after install

ianpittwood:
## Summary
- `install.sh` now deletes the downloaded `.deb`/`.rpm` after a successful install.
- Pre-existing installer files in the working directory are left untouched (the existing "don't re-download" signal doubles as "don't delete").
- Bumps `THIS_VERSION` to `1.2.0`, matching the precedent set when `-y`/`yes` was added.

## Test plan
- [x] `bash install.sh install <ver> -y` removes the downloaded package after install.
- [x] Pre-placing the package in the working directory and running install leaves the file in place.
- [x] `bash install.sh -h` reflects no new flags.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,8 +17,15 @@ fi
 # Run unattended; show no questions, assume default answers.
 # May also be set by the '-y'/'yes' options on the install action.
 RUN_UNATTENDED=${RUN_UNATTENDED:-0}
-if [[ "$3" == "-y" || "$3" == "yes" ]]; then
+if [[ "$3" == "-y" || "$3" == "yes" || "$4" == "-y" || "$4" == "yes" ]]; then
   RUN_UNATTENDED=1
+fi
+
+# Keep the downloaded installer file after a successful install.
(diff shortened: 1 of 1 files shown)


glin (approved):
👍 

I would say to omit the opt-out flag though, keep it simple, as I can't think of any good reason why you'd want to keep the installer afterwards. I doubt it'll break anyone either, as the quick install script isn't widely used or advertised (e.g. we don't show it on docs.posit.co)

ianpittwood:
> 👍
> 
> I would say to omit the opt-out flag though, keep it simple, as I can't think of any good reason why you'd want to keep the installer afterwards. I doubt it'll break anyone either, as the quick install script isn't widely used or advertised (e.g. we don't show it on docs.posit.co)

Works for me! I wasn't sure if you would prefer to keep the original functionality in tact. The flag has been removed.

glin (approved):

PR #310, opened 2026-05-30
Fix portable Rscript --version

owenjonesuob:
Fixes #309; see issue for full details.

PR #312, opened 2026-06-01
Parse trailing args correctly in portable `Rscript`

### LINKED ISSUE #311: Portable `Rscript` doesn't handle trailing args after `-e`
## Description

In the `portable-r` build, the compiled `Rscript` binary is replaced with a "relocatable wrapper". This wrapper aims to mimic the behaviour of `Rscript`.

If the wrapper is invoked with one or more `-e` arguments (i.e. expressions to evaluate), any trailing positional arguments are discarded.

## Example

```console
user@host:~$ Rscript -e 'print(commandArgs(trailing = TRUE))'
character(0)
user@host:~$
user@host:~$ Rscript -e 'print(commandArgs(trailing = TRUE))' ARG1=one ARG2=two
ARGUMENT 'ARG1=one' __ignored__

ARGUMENT 'ARG2=two' __ignored__

character(0)
user@host:~$
```

## Cause

The current implementation assumes that once `-e` has been seen, the remaining arguments can be passed directly to R:

https://github.com/rstudio/r-builds/blob/8139849bb24e2ad94738cbfdd9ee21379a18a7cd/builder/portable-r/make-relocatable.sh#L101-L102

https://github.com/rstudio/r-builds/blob/8139849bb24e2ad94738cbfdd9ee21379a18a7cd/builder/portable-r/make-relocatable.sh#L122-L123

However, `R` and `Rscript` differ with respect to how trailing args should be passed:

* `R` expects trailing args to follow a `--args` flag - more details here: https://stat.ethz.ch/R-manual/R-devel/doc/manual/R-intro.html#Invoking-R-from-the-command-line-1
* `Rscript` expects trailing args to simply follow the provided filepath or `-e` argument - docs here: https://stat.ethz.ch/R-manual/R-devel/doc/manual/R-intro.html#Invoking-R-from-the-command-line-1

So when at least one `-e` argument is present, the wrapper currently constructs a command that looks like `R -e "..." [args]`, rather than `R -e "..." --args [args]`. This leads to the `ARGUMENT  __ignored__` error seen above.

## Possible solution

Rather than passing args through to R as soon as we encounter an `-e` argument, I think we'll need to "collect" these arguments in a similar way to how we already collect `--options`, and then ultimately to construct an R command using everything we have collected... I'll open a PR attempting this!

## Example of a downstream consequence

Like #309, I first came across this when trying to run `future::plan(future::multisession)`. After applying the patch for #309, I was still unable to successfully run that expression - I instead saw lots of errors like the following:

```
`ARGUMENT 'MASTER=localhost' __ignored__
ARGUMENT 'PORT=11540' __ignored__
[...]

Error in workRSOCK() : PORT must be specified
Execution halted
[...]
```

Again this is from deep in the internals of {parallelly}, where [its attempt to launch a parallel worker](https://github.com/futureverse/parallelly/blob/6b38896d1b1441e66752b52aa2ab73dcbbc9243e/R/makeNodePSOCK.R#L878) using {parallel} fails because [the expected command args](https://github.com/wch/r-source/blob/cf5d0a0bb283f54d1c69438efe2c47dda0846da1/src/library/parallel/R/snowSOCK.R#L384) aren't found.

owenjonesuob:
Fixes #311.

Instead of passing all args following `-e` directly to R, we instead "collect" `-e` arguments and trailing args, similarly to how we are already collecting `--options`. Then all the collected arguments are used to construct the command which is ultimately executed.

Because we can't now use the "easy" option of `"$@"`, it becomes awkward to make sure all arguments are quoted correctly. I _think_ this implementation using bash arrays is behaving well, but I'd love to know if it could be improved!

---

I _think_ this implementation also now aligns better with how the compiled `Rscript` handles argument order. From [the docs](https://stat.ethz.ch/R-manual/R-devel/library/utils/html/Rscript.html):

> The prescribed order of arguments is important: e.g. `--verbose` specified after `-e` will be part of `args` and passed to the expression; the same will happen to `-e` specified after `file`.

To me, there's some slight ambiguity about what to do if there are multiple `-e` args _interspersed_ with trailing args. This seems like an unlikely situation to come across in real life, but you never know... for now, I've opted to collect _all_ `-e` args first, and then to collect anything else as a trailing arg. This could easily be adjusted though.

### DIFF
--- a/builder/portable-r/make-relocatable.sh
+++ b/builder/portable-r/make-relocatable.sh
@@ -87,39 +87,54 @@ r_opts=( "--slave" "--no-restore" )
 r_file=""
 r_args=()
 while [ $# -gt 0 ]; do
-  case "$1" in
-    --help|-h)
-      echo "Usage: Rscript [options] [-e expr [-e expr2 ...] | file] [args]"
-      exit 0
-      ;;
-    --version)
-      exec "${R_HOME}/bin/R" --slave -e 'cat(sprintf("Rscript (R) version %s.%s (%s)\n", R.version$major, R.version$minor, R.version$date))'
-      ;;
-    --default-packages=*)
-      R_DEFAULT_PACKAGES="${1#--default-packages=}"
-      export R_DEFAULT_PACKAGES
-      shift
-      ;;
-    -e)
-      # Collect R expressions - each -e is followed by an expression
+  if [ -n "$r_file" ]; then
+    # Everything after the filename is a trailing arg
+    r_args+=( "$1" )
+    shift
+  elif [ "${#r_exprs[@]}" -gt 0 ]; then
+    # Everything after the first -e arg, except for other -e args, is a trailing arg
+    if [ "$1" = "-e" ]; then
       r_exprs+=( "$1" "$2" )
       shift 2
-      ;;
-    --*)
-      # Collect R options (--verbose, --vanilla, --no-environ, etc.)
-      r_opts+=( "$1" )
+    else
+      r_args+=( "$1" )
       shift
-      ;;
-    *)
-      # Positional args - first one is the file to execute iff we don't have any -e expressions yet
-      if [ -z "$r_file" -a "${#r_exprs[@]}" -eq 0 ]; then
-        r_file="$1"
-      else
-        r_args+=( "$1" )
-      fi
-      shift
-      ;;
-  esac
+    fi
+  else
+    case "$1" in
+      --help|-h)
+        echo "Usage: Rscript [options] [-e expr [-e expr2 ...] | file] [args]"
+        exit 0
+        ;;
+      --version)
+        exec "${R_HOME}/bin/R" --slave -e 'cat(sprintf("Rscript (R) version %s.%s (%s)\n", R.version$major, R.version$minor, R.version$date))'
+        ;;
+      --default-packages=*)
+        R_DEFAULT_PACKAGES="${1#--default-packages=}"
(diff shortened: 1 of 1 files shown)


glin (commented):
Makes sense. For the ambiguity, I might peek at the Rscript source to see how it actually works, because you never know with docs 🙂 

Can you also add some unit tests for these cases e.g. https://github.com/rstudio/r-builds/pull/310/changes/8e602b0e4ae24d7684311b2aa3078b93b368f032?

builder/portable-r/make-relocatable.sh
```diff
@@ -82,55 +82,83 @@ export R_HOME
 # R understands --verbose, --vanilla, --no-environ, --no-site-file, etc.
 # directly, so we only need to handle --default-packages (env var conversion)
 # and file argument conversion (first positional arg -> --file=).
-r_opts=""
+r_exprs=()
+r_opts=( "--slave" "--no-restore" )
```
glin, line 86: I haven't verified, but going to assume bash arrays won't work in this `#!/bin/sh` script on all distros, especially Alpine/musllinux distros where bash isn't even installed by default.

owenjonesuob:
Superseded by #313

builder/portable-r/make-relocatable.sh
owenjonesuob: Oops yep I think you're probably right! I've taken a different approach in #313 which I think avoids this problem ☺️

PR #313, opened 2026-06-02
Parse trailing args correctly in portable `Rscript`

### LINKED ISSUE #311: Portable `Rscript` doesn't handle trailing args after `-e`

owenjonesuob:
Fixes #311, effectively by inserting `--args` into the list of variables `$@` which are subsequently passed to R.

---

Supersedes previous attempt in #312, which depended on bash arrays - I think this solution is shell-agnostic (and somewhat neater!).

### DIFF
--- a/builder/portable-r/test-manylinux.sh
+++ b/builder/portable-r/test-manylinux.sh
@@ -210,7 +210,7 @@ result=$("$RSCRIPT" -e 'cat("a")' -e 'cat("b")')
 echo "  Rscript -e -e: OK"
 
 # -e expressions with trailing arguments (regression test for #311)
-result=$("RSCRIPT" -e 'cat("one/"); cat("two three/")' -e 'cat(commandArgs(trailingOnly = TRUE), sep = " ")' ARG1 ARG2)
+result=$("$RSCRIPT" -e 'cat("one/"); cat("two three/")' -e 'cat(commandArgs(trailingOnly = TRUE), sep = " ")' ARG1 ARG2)
 [ "$result" = "one/two three/ARG1 ARG2" ] || { echo "FAIL: Rscript -e ARGS"; exit 1; }
 echo "  Rscript -e ARGS: OK"
 
--- a/builder/portable-r/test-manylinux.sh
+++ b/builder/portable-r/test-manylinux.sh
@@ -209,6 +209,11 @@ result=$("$RSCRIPT" -e 'cat("a")' -e 'cat("b")')
 [ "$result" = "ab" ] || { echo "FAIL: Rscript -e -e"; exit 1; }
 echo "  Rscript -e -e: OK"
 
+# -e expressions with trailing arguments (regression test for #311)
+result=$("RSCRIPT" -e 'cat("one/"); cat("two three/")' -e 'cat(commandArgs(trailingOnly = TRUE), sep = " ")' ARG1 ARG2)
+[ "$result" = "one/two three/ARG1 ARG2" ] || { echo "FAIL: Rscript -e ARGS"; exit 1; }
+echo "  Rscript -e ARGS: OK"
+
 # File execution with arguments
 tmpscript=$(mktemp /tmp/test_rscript_XXXXXX.R)
 echo 'args <- commandArgs(trailingOnly=TRUE); cat(paste(args, collapse=","))' > "$tmpscript"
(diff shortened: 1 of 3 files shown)


owenjonesuob:
NB. I had a snoop at the [Rscript source code](https://github.com/wch/r-source/blob/cf5d0a0bb283f54d1c69438efe2c47dda0846da1/src/unix/Rscript.c#L302) - I interpret it as:

1. If we see `-e`, then activate "e_mode" and interpret the following arg as an expression
2. As soon as we don't see `-e`, then stop looking for expressions, everything else is a trailing arg

I've aimed for the same logic with the changes in this PR 🙂

builder/portable-r/test-manylinux.sh
```diff
@@ -209,6 +209,11 @@ result=$("$RSCRIPT" -e 'cat("a")' -e 'cat("b")')
 [ "$result" = "ab" ] || { echo "FAIL: Rscript -e -e"; exit 1; }
 echo "  Rscript -e -e: OK"
 
+# -e expressions with trailing arguments (regression test for #311)
+result=$("RSCRIPT" -e 'cat("one/"); cat("two three/")' -e 'cat(commandArgs(trailingOnly = TRUE), sep = " ")' ARG1 ARG2)
```
glin: There's just one test issue here I noticed, otherwise looks great. I don't know why I can't approve this PR to run the CI checks, but will try to get that working in the future.

```sh
++ RSCRIPT -e 'cat("one/"); cat("two three/")' -e 'cat(commandArgs(trailingOnly = TRUE), sep = " ")' ARG1 ARG2
/r-builds/builder/portable-r/test-manylinux.sh: line 213: RSCRIPT: command not found
+ result=

make: *** [Makefile:40: test-r-manylinux_2_34] Error 127
```

typo with `RSCRIPT` -> `$RSCRIPT`?
(and ditto for the musllinux test)

owenjonesuob: Oops yep - fixed in 6c6e20e!

glin (approved):
Great, thanks!

PR #317, opened 2026-06-17
Portable Linux: replace `Rscript` shell script with a native `Rscript` wrapper using `RHOME`

### LINKED ISSUE #315: portable linux: keep the original Rscript binary
Unfortunately Rscript behaves differently from the current replacement shell script, e.g. the shell script starts an interactive R shell if I call it without arguments, instead of printing the help like the original Rscript binary does. 

I am sure that there are other differences as well, and the differences might change over R versions, so in general we probably don't want to track these differences.

Why not simplify the shell script to simply setting `RHOME` and calling the original Rscript binary? That should work for all R versions we care about.

Something like this should do, this is what I use on macOS:
```
#!/bin/sh
RHOME=$(cd "$(dirname "$(realpath "$0")")/.." && pwd -P)
export RHOME
exec "$(dirname "$(realpath "$0")")/Rscript.orig" "$@"
```

Is there anything against this?

glin:
Closes #315 

@gaborcsardi pointed out that the native `Rscript` reads an `RHOME` env var that we can just set in a wrapper script. It's much simpler than reimplementing Rscript, especially with all the small bugs we've had to fix so far. The portable macOS R build also does this same thing.

I couldn't find a good reason for reimplementing `Rscript` in the original work besides just not knowing that `RHOME` was configurable. It's even in the Rscript docs:
> Normally the version of R is determined at installation, but this can be overridden by setting the environment variable RHOME.
> https://stat.ethz.ch/R-manual/R-devel/library/utils/html/Rscript.html

I tested this on various distros and R versions going back to 3.0.0 and it seems to work perfectly fine. The `Rscript` wrapper scripts look like this (`Rscript.bin` is the original `Rscript` moved):

```sh
$ cat /opt/R/4.6.0/bin/Rscript
#!/bin/sh
# Relocatable Rscript wrapper: set RHOME from our own location, then exec the
# preserved native binary, which reads RHOME and execs "${RHOME}/bin/R".
RHOME="$(dirname "$(dirname "$(readlink -f "$0")")")/lib/R"
export RHOME
exec "$(dirname "$(readlink -f "$0")")/Rscript.bin" "$@"

$ cat /opt/R/4.6.0/lib/R/bin/Rscript
#!/bin/sh
# Relocatable Rscript wrapper: set RHOME from our own location, then exec the
# preserved native binary, which reads RHOME and execs "${RHOME}/bin/R".
RHOME="$(dirname "$(dirname "$(readlink -f "$0")")")"
export RHOME
exec "$(dirname "$(readlink -f "$0")")/Rscript.bin" "$@"
```

### DIFF
--- a/builder/portable-r/make-relocatable.sh
+++ b/builder/portable-r/make-relocatable.sh
@@ -55,101 +55,42 @@ R_HOME_DIR="$(dirname "$(dirname "$(readlink -f "$0")")")"
   echo "  Patched lib/R/bin/R"
 fi
 
-echo ">>> Replace Rscript with relocatable wrapper"
-# Replace compiled Rscript binary with a relocatable shell wrapper.
-# The compiled Rscript has R_HOME hardcoded at compile time and doesn't check
-# the R_HOME env var, so it can't find bin/R after relocation.
-# The wrapper replicates Rscript's argument handling: options like --verbose,
-# --default-packages are processed; -e expressions pass through; the first
-# positional arg becomes --file=arg; remaining positional args use --args.
+echo ">>> Make Rscript relocatable (preserve binary, wrap with RHOME)"
+# The compiled Rscript binary has R_HOME baked in at compile time, so after
+# relocation it would exec the wrong (build-time) bin/R.
+#
+# But Rscript reads the RHOME environment variable and, when set, execs
+# "${RHOME}/bin/R" instead of the compiled-in path. This has held for every R
+# version we support (3.0+). See getenv("RHOME") in R's src/unix/Rscript.c:
+# https://github.com/wch/r-source/blob/tags/R-4-6-0/src/unix/Rscript.c#L230-L280
+#
+# So we preserve the original binary as Rscript.bin and install a thin wrapper
+# that sets RHOME from its own location and execs the real binary.
+#
+# Rscript depends only on libc + ld (no bundled libs), so the preserved binary
+# needs no RPATH/patchelf handling.
 for rscript in "${R_INSTALL_PATH}/bin/Rscript" "${R_INSTALL_PATH}/lib/R/bin/Rscript"; do
   [ -f "$rscript" ] || continue
-  # Determine R_HOME formula based on location (same logic as bin/R vs lib/R/bin/R)
+  # RHOME must be R_HOME (<prefix>/lib/R). Formula depends on the script's depth,
+  # same logic as the bin/R vs lib/R/bin/R distinction above.
+  #   bin/Rscript        is at <prefix>/bin/Rscript        -> RHOME=<prefix>/lib/R
+  #   lib/R/bin/Rscript  is at <R_HOME>/bin/Rscript        -> RHOME=<R_HOME>
   if [[ "$rscript" == */lib/R/bin/Rscript ]]; then
-    r_home_formula='R_HOME="$(dirname "$(dirname "$(readlink -f "$0")")")"'
+    rhome_formula='RHOME="$(dirname "$(dirname "$(readlink -f "$0")")")"'
   else
-    r_home_formula='R_HOME="$(dirname "$(dirname "$(readlink -f "$0")")")/lib/R"'
+    rhome_formula='RHOME="$(dirname "$(dirname "$(readlink -f "$0")")")/lib/R"'
   fi
-  cat > "$rscript" <<'RSCRIPT_OUTER'
+  mv "$rscript" "${rscript}.bin"
+  cat > "$rscript" <<'RSCRIPT_HEAD'
 #!/bin/sh
-# Relocatable Rscript wrapper (replaces compiled binary for portability)
-RSCRIPT_OUTER
-  echo "${r_home_formula}" >> "$rscript"
-  cat >> "$rscript" <<'RSCRIPT_EOF'
-export R_HOME
-
-# Process Rscript-specific options, collect pass-through R options.
-# R understands --verbose, --vanilla, --no-environ, --no-site-file, etc.
-# directly, so we only need to handle --default-packages (env var conversion)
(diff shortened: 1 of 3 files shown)


gaborcsardi (approved):

glin:
@gaborcsardi thanks again! Portable R binaries have been rebuilt on all patch versions.

There was one issue though: apparently there's a bug in R 3.4.x where `Rscript` segfaults on musl specifically in some cases. I'm looking more into whether there's a simple workaround for it, but otherwise it's probably ok to ignore for an old EOL R version.
```sh
+ /opt/R/3.4.4/lib/R/bin/Rscript /r-builds/test/test.R
Installing package into ‘/tmp/RtmphgIPnk’
(as ‘lib’ is unspecified)

 *** caught segfault ***
address 0x7f1d9f8e27c8, cause 'memory not mapped'

Traceback:
 1: download.file(url = paste0(repos, "/PACKAGES.rds"), destfile = dest,     method = method, cacheOK = FALSE, quiet = TRUE, mode = "wb")
 2: doTryCatch(return(expr), name, parentenv, handler)
 3: tryCatchOne(expr, names, parentenv, handlers[[1L]])
 4: tryCatchList(expr, classes, parentenv, handlers)
 5: tryCatch({    download.file(url = paste0(repos, "/PACKAGES.rds"), destfile = dest,         method = method, cacheOK = FALSE, quiet = TRUE, mode = "wb")}, error = identity)
 6: available.packages(contriburl = contriburl, method = method)
 7: install.packages("pkgconfig")
An irrecoverable exception occurred. R is aborting now ...
/r-builds/builder/portable-r/../../test/test-r.sh: line 18:   141 Segmentation fault      (core dumped) DIR=$SCRIPT_DIR "${R_HOME}/bin/Rscript" "${SCRIPT_DIR}/test.R"
```

gaborcsardi:
Yeah, that's actually a bug in R. E.g. I have this in`rhub/r-minimal`:

```
docker run -it rhub/r-minimal:3.4 bash
```

```
R -q -e 'install.packages("pkgconfig", repos = "https://cloud.r-project.org")'
```

```
> install.packages("pkgconfig", repos = "https://cloud.r-project.org")
trying URL 'https://cloud.r-project.org/src/contrib/pkgconfig_2.0.3.tar.gz'
Content type 'application/x-gzip' length 6080 bytes
==================================================
downloaded 6080 bytes

* installing *source* package ‘pkgconfig’ ...
** package ‘pkgconfig’ successfully unpacked and MD5 sums checked
** R
** inst
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded
* DONE (pkgconfig)

The downloaded source packages are in
	‘/tmp/RtmpGeEOCJ/downloaded_packages’
Updating HTML index of packages in '.Library'
Making 'packages.html' ... done
```

But with Rscript:

```
Rscript -e 'install.packages("pkgconfig", repos = "https://cloud.r-project.org")'
```

```
trying URL 'https://cloud.r-project.org/src/contrib/pkgconfig_2.0.3.tar.gz'
Content type 'application/x-gzip' length 6080 bytes
==================================================
downloaded 6080 bytes


 *** caught segfault ***
address 0xffffa4448438, cause 'memory not mapped'

Traceback:
 1: download.file(url, destfile, method, mode = "wb", ...)
 2: doTryCatch(return(expr), name, parentenv, handler)
 3: tryCatchOne(expr, names, parentenv, handlers[[1L]])
 4: tryCatchList(expr, classes, parentenv, handlers)
 5: tryCatch(expr, error = function(e) {    call <- conditionCall(e)    if (!is.null(call)) {        if (identical(call[[1L]], quote(doTryCatch)))             call <- sys.call(-4L)        dcall <- deparse(call)[1L]        prefix <- paste("Error in", dcall, ": ")        LONG <- 75L        msg <- conditionMessage(e)        sm <- strsplit(msg, "\n")[[1L]]        w <- 14L + nchar(dcall, type = "w") + nchar(sm[1L], type = "w")        if (is.na(w))             w <- 14L + nchar(dcall, type = "b") + nchar(sm[1L],                 type = "b")        if (w > LONG)             prefix <- paste0(prefix, "\n  ")    }    else prefix <- "Error : "    msg <- paste0(prefix, conditionMessage(e), "\n")    .Internal(seterrmessage(msg[1L]))    if (!silent && identical(getOption("show.error.messages"),         TRUE)) {        cat(msg, file = outFile)        .Internal(printDeferredWarnings())    }    invisible(structure(msg, class = "try-error", condition = e))})
 6: try(download.file(url, destfile, method, mode = "wb", ...))
 7: download.packages(pkgs, destdir = tmpd, available = available,     contriburl = contriburl, method = method, type = "source",     quiet = quiet, ...)
 8: install.packages("pkgconfig", repos = "https://cloud.r-project.org")
An irrecoverable exception occurred. R is aborting now ...
Segmentation fault (core dumped)
```

It happens because Rscript used to load a different set of default packages than R. Look for `R_SCRIPT_DEFAULT_PACKAGES` and `R_SCRIPT_LEGACY` in the R news and the source code. (Btw. this is yet another (former) difference between R and Rscript.) Old Rscript did not load the 'methods' package. (Yes, this was the most annoying "feature" in R ever, and I can't believe that I am still dealing with it about 10 years after it was fixed.)

E.g. this works:
```
Rscript -e 'library(methods); install.packages("pkgconfig", repos = "https://cloud.r-project.org")'
```

IDK if we want to reproduce the crash or not. Maybe yes. Then the workaround is to load the 'methods' package in the `test.R` file. On newer R it is loaded already, anyway.

glin:
@gaborcsardi I see, and yeah I agree that it makes sense to leave it alone and just fix the tests. Especially since it's an old EOL R anyway.

PR #320, opened 2026-06-22
Load methods in older Rscript, to match R

gaborcsardi:
Fixes tests on musllinux-1.2 on R 3.4.x.
See https://github.com/rstudio/r-builds/pull/317#issuecomment-4749554642

### DIFF
--- a/test/test.R
+++ b/test/test.R
@@ -1,3 +1,6 @@
+# Older Rscript did not load the methods package by default
+library(methods)
+
 # HTTP mirror to support R 3.1
 options(repos = c("https://cloud.r-project.org", "http://cloud.r-project.org"))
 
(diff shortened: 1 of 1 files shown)


test/test.R
```diff
@@ -1,3 +1,6 @@
+# Older Rscript did not load the methods package by default
```
glin, line 1: Might be useful to document here that this specifically prevents an R 3.4.x bug from crashing Rscript on musllinux. Otherwise 👍

glin (approved):

test/test.R
gaborcsardi: OK. I thought people would look at the commit message. :) But let me add a comment.

gaborcsardi: done.

glin (approved):

PR #321, opened 2026-06-23
Skip libcurl tests for R 3.4.x musllinux for now to work around Rscript seg fault

glin:
Continued from https://github.com/rstudio/r-builds/pull/320, `library(methods)` worked for R 3.4.2 to 3.4.4, but unfortunately not for 3.4.0/3.4.1.

So let's just take the easy way out and skip the libcurl tests to unblock the musllinux builds. The rest of the tests continue to run for R 3.4.x.


gaborcsardi (approved):
Thanks!

PR #327, opened 2026-06-30
Add Cloudsmith retention for nightly R-next/R-devel builds

jstruzik:
## Problem

The daily devel/next builds (`devel-daily.yml`) publish a new **date-versioned** `R-next`/`R-devel` package for every platform × architecture, every night. Nothing prunes them, so `posit/open` grows without bound (~50 new package versions/day). This is the storage side of the broader Cloudsmith-usage concern (cf. [CLP-676](https://positpbc.atlassian.net/browse/CLP-676)).

## Change

- New `prune_cloudsmith.py`: keeps the newest **14** versions of each nightly package *coordinate* (name + distribution + architecture) and deletes the rest, via the cloudsmith CLI.
- New `prune-cloudsmith` job in `devel-daily.yml`, run after the build/publish completes.
- Selection logic is pure and unit-tested in `test_prune_cloudsmith.py` (wired into `make unit-test`).

## Safety properties

- **Count-based, not age-based** — keeping newest-N per coordinate means an outage or a run of failed builds can never prune a coordinate to zero (age-based could, if nothing fresh replaces what ages out).
- **Gated on success** — `if: success() && ... && cloudsmith_dry_run != true`, so a failed nightly never triggers deletions, and dry runs don't prune.
- **Nightlies only** — only `R-next`/`R-devel` (and deb-cased `r-next`/`r-devel`) are eligible; stable release packages (`R-4.4.3`, …) are never deletion candidates, verified by a test.

## Validation note

The selection logic is fully unit-tested. The cloudsmith CLI list/delete plumbing (JSON field names, delete identifier) could not be exercised against the live API here — worth a one-time check of the first run's logs, or a manual `python prune_cloudsmith.py --repo posit/open --keep 14 --dry-run`, to confirm field mapping before relying on it. The script fails closed (lists nothing → deletes nothing).

Depends on the `CLOUDSMITH_API_KEY` token having a delete entitlement on `posit/open`.


jstruzik:
We can see a successful dry run of the pruning here: https://github.com/rstudio/r-builds/actions/runs/28475898530/job/84400285289

glin (approved):

PR #331, opened 2026-07-17
Handle arm64 architecture in the quick install script

glin:
`download_name` only recognized `aarch64`, so an `arm64` arch string (or any unknown arch) built a bad package URL that 404s instead of failing with a clear error. This maps `arm64` to the same packages, fails fast on anything unsupported, and adds offline unit tests so the arch logic is covered in CI without an arm64 host.

Fixes an issue reported by @bschwedler about arm64 PPM images failing to install R properly via the quick install script.

---

## Changes

- Map `arm64` as an alias for `aarch64`, covering both the deb and rpm package names.
- Fail fast when there's nothing to install, reporting an unsupported architecture and an unsupported OS (e.g. an Ubuntu derivative) separately instead of building a 404 URL.
- Make `install.sh` sourceable: move arg/env parsing and dispatch into `main()`, the `versions.json` fetch into `load_r_versions()`, and guard execution so curl-pipe, piped, and direct runs are unchanged while sourcing loads only the functions.
- Add `test/test-install-unit.sh`: 50 offline assertions (both arches, both package types, the failure paths, and the source guard), wired into `make unit-test`.
- Bump the script version to 1.3.0.

## Testing

- `make unit-test`: 50 passed.
- End-to-end on emulated arm64 via the curl-pipe form: Ubuntu 24.04 installs the arm64 deb, Rocky 9 installs the aarch64 rpm, R 4.6.1 runs in both (exit 0).
- arm64 packages confirmed on the CDN for ubuntu-2404 (deb) and rhel-9/10, opensuse-156, fedora-42, centos-7 (rpm).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,30 +1,13 @@
 #!/bin/bash
-THIS_VERSION="1.2.0"
+THIS_VERSION="1.3.0"
 
 # Call with:
 #   bash -c "$(curl -L https://rstd.io/r-install)"
-
-SCRIPT_ACTION=$1
-SCRIPT_ACTION=${SCRIPT_ACTION:-install}
-
-# Set to the full version to install. Must be either available on S3 or in the working directory
-R_VERSION=${R_VERSION:-}
-# The version may optionally be provided as a second argument
-if [[ "$2" != "" ]]; then
-  R_VERSION=$2
-fi
-
-# Run unattended; show no questions, assume default answers.
-# May also be set by the '-y'/'yes' options on the install action.
-RUN_UNATTENDED=${RUN_UNATTENDED:-0}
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -163,10 +150,17 @@ download_name () {
       rpm_arch="x86_64"
       deb_arch="amd64"
       ;;
-    "aarch64")
+    # Linux reports 64-bit ARM as "aarch64"; "arm64" is accepted as an alias
+    # (e.g., dpkg's architecture name) so the mapping never silently misses.
+    "aarch64" | "arm64")
```
glin, line 155: This is the key change to review.

Everything else is just adding much needed tests for the script + refactoring the script to be testable.

bschwedler (approved):
Thanks Greg!

PR #337, opened 2026-07-21
Fix R 4.2.2 build on Fedora 44

gaborcsardi:
Apparently, this R version on this Fedora version needs an extra (la)tex package. I tested this locally.

### DIFF
--- a/builder/Dockerfile.fedora-44
+++ b/builder/Dockerfile.fedora-44
@@ -35,7 +35,7 @@ RUN dnf -y upgrade \
     tex \
     texinfo-tex \
     texlive-collection-latexrecommended \
-    texlive-rsfs \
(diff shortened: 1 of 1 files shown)


builder/Dockerfile.fedora-44
```diff
@@ -35,6 +35,7 @@ RUN dnf -y upgrade \
     tex \
     texinfo-tex \
     texlive-collection-latexrecommended \
+    texlive-rsfs \
```
glin: A comment would be useful here just because this is so odd. Some other dockerfiles do inline comments like https://github.com/rstudio/r-builds/blob/e4291be9d92eb9c435504a3cddf02fa6f381c247/builder/Dockerfile.manylinux_2_34#L30

glin (approved):

